Issue #16, opened 2025-07-08
refactor: Streamline Chat Bubble Props and Message Types for Consistency


aryanranderiya:
## Description

The chat bubble components currently have inconsistent prop interfaces and scattered message type definitions across the codebase. This creates maintenance overhead, potential bugs due to type mismatches, and makes it difficult to extend bubble functionality consistently.

### Inconsistent Prop Interfaces

**Different bubble components use varying prop structure patterns:**
- `ChatBubbleUserProps`: 18 properties (message_id, text, searchWeb, fileData, etc.)
- `ChatBubbleBotProps`: 41+ properties (includes all user props plus tool-specific data)
- `ChatBubbleActionsProps`: 4 properties (loading, text, pinned, message_id)  
- `ChatBubbleActionsImageProps`: 3 properties (fullWidth, setOpenImage, image_data)

**Tool-specific bubble sections have their own interfaces:**
- `TodoSectionProps`: todos, projects, stats, action, message
- `GoalSectionProps`: goals, stats, action, message, goal_id, error
- `SearchResultsProps`: Varies by result type (news, web, images, videos)

### Message Type Inconsistencies

**Multiple message type definitions exist across different files:**
- `MessageType` in `convoTypes.ts` (main structure with ~25 fields)
- `MessageModel` in backend `chat_models.py` (similar but different structure)
- `Message` in `onboarding/types` (simple: id, type, content)
- `Message` in `browser/page.tsx` (role, content, stepData)
- `StreamChunk` and partial message types in various hooks

### Scattered Type Definitions

**Props are not standardized across bubble types:**
- Some components receive 40+ individual props via destructuring
- Others use focused, minimal interfaces
- No inheritance hierarchy or base interface structure
- Mixed patterns for handling tool data (direct props vs nested objects)

## Impact

- **Maintenance overhead**: Adding new message fields requires updating multiple scattered interfaces
- **Type safety issues**: Different bubble components expect data in different formats
- **Development complexity**: No consistent pattern for extending bubble functionality
- **Error-prone**: Large, complex prop interfaces increase likelihood of mistakes
- **Scalability problems**: Difficult to add new bubble types or modify existing ones

### 1. Standardized Base Interfaces
Create a hierarchy of base interfaces that all bubble components extend:
```typescript
interface BaseBubbleProps {
  message_id: string;
  date?: string;
  loading?: boolean;
}

interface BaseChatBubbleProps extends BaseBubbleProps {
  text: string;
[...]

PR #59, opened 2025-07-14
refactor: chat service and message handling logic

aryanranderiya:
Streamline data extraction and type management for chat messages, simplify content checks, and enhance utility functions for better code organization and maintainability.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,12 +11,17 @@ repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v5.0.0
     hooks:
-      - id: trailing-whitespace # Removes trailing whitespace on all lines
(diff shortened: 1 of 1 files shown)


aryanranderiya:
Closes #16

aryanranderiya:
@Dhruv-Maradiya please check messageContentUtils.ts and replace hardcoded fields with dynamic field extraction from the message or chat types

backend/app/api/v1/router/websocket.py
Dhruv-Maradiya, line 1: Is this comment intended?

aryanranderiya: Yes it's to ignore pre commit error of 'bandit' where Exception: pass isn't allowed because exceptions should be raised, but in this case if the error occurs in the websocket we just want to ignore it so pass is acceptable, the comment line is to suppress the error in the backend build pipeline

PR #80, opened 2025-07-30
 Fix: Prevent Duplicate Error Toasts on Unauthorized Requests

Vinitthakkar45:
This PR resolves the issue of duplicate error toasts appearing when API requests fail due to unauthorized access (401).
 or duplicate failure or success msgs

### DIFF
--- a/backend/app/utils/notification/orchestrator.py
+++ b/backend/app/utils/notification/orchestrator.py
@@ -263,7 +263,9 @@ class NotificationOrchestrator:
             await self.storage.update_notification(
                 notification_id, result.update_notification
             )
-            logger.info(f"Broadcasting notification {notification.id} to user {notification.user_id}")
+            logger.info(
+                f"Broadcasting notification {notification.id} to user {notification.user_id}"
+            )
             # Broadcast update to user via websocket
             await websocket_manager.broadcast_to_user(
                 user_id,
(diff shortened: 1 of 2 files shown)


backend/app/utils/notification/orchestrator.py
```diff
@@ -263,7 +263,9 @@ async def execute_action(
             await self.storage.update_notification(
                 notification_id, result.update_notification
             )
-
+            logger.info(
+                f"Broadcasting notification {notification.id} to user {notification.user_id}"
+            )
```
aryanranderiya, line 268:
```suggestion
await self.storage.update_notification(
notification_id, result.update_notification
)
```

frontend/src/features/calendar/api/calendarApi.ts
aryanranderiya, line 1: Why are we removing the error message when unable to fetch calendars?

frontend/src/features/mail/api/mailApi.ts
aryanranderiya, line 1: Why are we removing the error message when unable to fetch mails? (Same as calendar, we can remove the duplication of the normal toast instead of the error message). This is trivial, so will ignore it along with ignoring the calendar change, but let me know if there's some specific reason

aryanranderiya (changes requested):

frontend/src/features/mail/api/mailApi.ts
Vinitthakkar45: there were flooding error messages of "failed to fetch details" as frontend is not currently checking if user has connected before fetching data from backend

aryanranderiya (approved):

PR #105, opened 2025-08-09
Enhance OAuth integration with unified support and caching

aryanranderiya:
Improve handling of OAuth integrations by introducing support for unified integrations and implementing caching for integrations configuration to enhance performance.


aryanranderiya:
We need to look into how choosing specific files works for the drive.file oauth scope, because we switched from drive scope to drive.file so need to implement file picking.

Dhruv-Maradiya (approved):

PR #206, opened 2025-09-06
fix: setup.sh for cross-platform usage:

jigarbhoye04:
- Detects python3 or python automatically
- Activates virtual environment on Windows and Unix
- Installs uv inside the venv if not already installed
- Ensures backend dependencies install correctly across OSes


jigarbhoye04:
I've verified that this is now working properly on a Windows machine.

please, check once on you system if it's working properly or not.

Dhruv-Maradiya:
Hey @jigarbhoye04 I tried running the script and encountered few issues.

- Shell compatibility: Script fails with source: not found when run with sh Needed: Replace source with . for POSIX compliance
- Missing pip in venv: I dont know why but I face this issue constantly, Virtual environment created without pip module
- Verify Python and pip are working in venv

so I made few changes to the script. Can you please verify if its working on the Windows or not.

Thank you for your contribution.

aryanranderiya (approved):

PR #219, opened 2025-09-16
feat(backend): add HEALTHCHECK to Dockerfile

tansugangopadhyay:
Hi, I've added a HEALTHCHECK instruction to the backend/Dockerfile. This will help improve the reliability of the application in production by allowing Docker to monitor its health.

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -93,5 +93,7 @@ COPY --chown=appuser:appuser . .
 USER appuser
 EXPOSE 8000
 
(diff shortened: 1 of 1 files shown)


aryanranderiya:
Hey, thank you for the contribution! Thanks for adding this. We already have a healthcheck defined in docker-compose.yml, so adding another one in the Dockerfile would duplicate the check and could lead to drift between the two. Since we’re currently relying on Compose, I’d suggest removing the HEALTHCHECK from the Dockerfile and keeping the Compose one as the source of truth.


CleanShot 2025-09-22 at 15 27 21@2x

PR #240, opened 2025-09-26
feat(chat): Implement Dexie.js Caching

jigarbhoye04:
This PR closes (https://github.com/heygaia/gaia/issues/122) and implements a local-first caching architecture using Dexie.js to reduce chat latency.

The chat feature has been refactored for improved performance and responsiveness. Chat data now loads instantly from a local IndexedDB cache, with network operations (fetching, sending, deleting) handled in the background.

### key features added

* Instant Loading: Chat list and message history load immediately from the local cache.
* Updates: Sending and deleting messages update the UI instantly while network requests complete in the background.
* Offline: Users can view cached conversations offline. Failed actions are handled correctly with UI rollbacks.
* Caching: Background sync fetches messages for server-initiated conversations, ensuring zero-latency loading for new chats.

### architectural changes

- Persistence (`chatDb.ts`): Dexie.js service encapsulating IndexedDB logic, schemas, typed models, and atomic transactions.
- State & Logic (`chatStore.ts` & Hooks):
   central Zustand store manages UI-related chat state.
   cache-first hooks (`useConversations`, `useMessages`) orchestrate DB -> UI -> Network flow.
   action hooks (`useSendMessage`, `useDeleteConversation`) handle user writes.
- UI (`ChatsList.tsx`, `ChatPage.tsx`, etc.):

- UI components consume hooks declaratively, removing complex local state and data-fetching logic.


frontend/src/services/syncService.ts
```diff
@@ -0,0 +1,113 @@
...
+    if (remoteConversationIds.length === 0) {
+      return;
+    }
+
+    const cachedSet = new Set(cachedConversationIds);
+    const pendingConversationIds = remoteConversationIds.filter(
+      (conversationId) => !cachedSet.has(conversationId),
+    );
+
+    for (const conversationId of pendingConversationIds) {
```
Dhruv-Maradiya: Its better practice to parallelize these api calls into `Promise.all` or `Promise.allResolved` and handle exceptions.

frontend/src/services/syncService.ts
```diff
@@ -0,0 +1,113 @@
...
+    case "user":
+      return "user";
+    case "bot":
+      return "assistant";
+    default:
+      return "system";
+  }
+};
+
+const fetchAllConversationIds = async (): Promise<string[]> => {
```
Dhruv-Maradiya: I think we should not fetch all the conversations, 
Don’t let IndexedDB grow indefinitely. For example, keep only the last N conversation.

frontend/src/hooks/useDeleteConversation.ts
```diff
@@ -0,0 +1,64 @@
...
+      const snapshot = useChatStore.getState();
+      const conversation: IConversation | undefined = snapshot.conversations.find(
+        (item) => item.id === conversationId,
+      );
+      const messages: IMessage[] =
+        snapshot.messagesByConversation[conversationId] ?? [];
+
+      removeConversation(conversationId);
+
+      try {
```
Dhruv-Maradiya: Here we should parallelize using `Promise.allResolved` and exceptions can be handled as well.

frontend/src/features/chat/api/chatApi.ts
```diff
@@ -166,6 +172,33 @@ export const chatApi = {
     );
   },
 
+  // Send a new message to a conversation
```
Dhruv-Maradiya: `sendMessage` is not needed, its handled from the backend

frontend/src/features/chat/components/composer/Composer.tsx
```diff
@@ -112,7 +114,7 @@ const Composer: React.FC<MainSearchbarProps> = ({
     clearSelectedWorkflow();
 
     setIsLoading(true);
-    sendMessage("Run this workflow", [], null, null, selectedWorkflow);
+  sendMessage("Run this workflow", conversationId ?? "");
```
Dhruv-Maradiya: There's some formatting issue

frontend/src/hooks/useConversations.ts
```diff
@@ -0,0 +1,92 @@
...
+        if (isActive) {
+          setConversations(cachedConversations);
+        }
+      } catch {
+        // Ignore cache read errors and proceed with network fetch
+      }
+
+      try {
+        const { conversations: apiConversations } =
+          await chatApi.fetchConversations(1, 20);
```
Dhruv-Maradiya, line 45: Can you please explain why `useConversations` fetches conversation as well as load from indexdb, how flow is working

frontend/src/services/syncService.ts
```diff
@@ -0,0 +1,113 @@
...
+      try {
+        const messages = await chatApi.fetchMessages(conversationId);
+        if (messages.length === 0) continue;
+
+        const mappedMessages = mapApiMessagesToStored(
+          messages,
+          conversationId,
+        );
+
+        if (mappedMessages.length === 0) continue;
```
Dhruv-Maradiya: This check is not needed. As mappedMessages will never be empty

frontend/src/hooks/useConversations.ts
jigarbhoye04: We hydrate from IndexedDB first so the sidebar populates instantly. Once that is done, we then fetch page 1 from the API to guarantee fresh data, update dexie with the latest payload, and replace the in-memory list. that keeps the UI responsive while staying in sync with the server.

Dhruv-Maradiya: Yeah right, but there's hook registered in layout named `useBackgroundSync` that does sync data with backend right?

jigarbhoye04:
Background sync now processes only the latest 50 conversations from the server, skips already-fresh threads, and pre-caches messages in parallel while updating their metadata. Sending a message no longer uses the redundant REST endpoint, and delete operations run API/DB work concurrently with clean rollback handling.

frontend/src/hooks/useConversations.ts
jigarbhoye04: `useBackgroundSync` comes into play when the app is up and running - it quietly preloads messages for any conversation the server might have so that actually opening them feels instant. But on the first load of the app, `useConversations` still has its way of getting whatever is already in IndexedDB out in the open immediately and then fetching the first page from the API to refresh titles, stars, and metadata. Afterward, background sync carries on making the messages available from the background - it never replaces that initial conversation load.

Dhruv-Maradiya: Okay, make sense

Dhruv-Maradiya:
Hey @jigarbhoye04 please resolve all conversation when you think its ready to be reviewed again. Thanks.

frontend/src/features/chat/components/composer/Composer.tsx
```diff
@@ -112,7 +114,7 @@ const Composer: React.FC<MainSearchbarProps> = ({
     clearSelectedWorkflow();
 
     setIsLoading(true);
-    sendMessage("Run this workflow", [], null, null, selectedWorkflow);
+    sendMessage("Run this workflow", conversationId ?? "");
```
Dhruv-Maradiya: selectedWorkflow must be sent to the backend so it can be executed. Sending only conversationId won't work.

frontend/src/features/chat/components/composer/Composer.tsx
```diff
@@ -192,13 +194,7 @@ const Composer: React.FC<MainSearchbarProps> = ({
...
     setContextualLoading(true, inputText);
 
-    sendMessage(
-      inputText,
-      uploadedFileData,
-      selectedTool,
-      selectedToolCategory,
-      selectedWorkflow,
-    );
+    sendMessage(inputText, conversationId ?? "");
```
Dhruv-Maradiya: Same pattern as above, selectedWorkflow, selectedToolCategory, selectedTool and uploadedFileData must be sent.

Dhruv-Maradiya (changes requested):

frontend/src/services/syncService.ts
```diff
@@ -0,0 +1,155 @@
...
+      MAX_SYNC_CONVERSATIONS,
+    );
+    return conversations;
+  } catch {
+    return [];
+  }
+};
+
+const mapApiConversation = (conversation: Conversation): IConversation => ({
+  id: conversation.conversation_id,
```
aryanranderiya: what is the need for this new Interface "IConversation" ? 
Why is the mapping and new type needed? Can't we use Conversation type itself, and if we need extra props we can simply extend from the existing?

frontend/src/lib/db/chatDb.ts
```diff
@@ -0,0 +1,126 @@
+import Dexie, { Table } from "dexie";
+
+import { SystemPurpose } from "@/features/chat/api/chatApi";
+import { FileData } from "@/types/shared";
+
+type MessageRole = "user" | "assistant" | "system";
+type MessageStatus = "sending" | "sent" | "failed";
+
+export interface IConversation {
```
aryanranderiya, line 9: Possibly redundant type. Refer to previous comment.

jigarbhoye04: we are still depending on IConversation for the Dexie schema, however, there is now a sole mapper that renders support to the hook as well as the background sync.

frontend/src/services/syncService.ts
jigarbhoye04: The mapping logic has been completely moved into `useConversations` and it has been reused from sync service so there is no more duplication. We still maintain `IConversation` because the stores of Dexie work on that normalized shape (string id, date instances, optional fields flattened for IndexedDB).

frontend/src/features/chat/components/composer/Composer.tsx
jigarbhoye04: Right, sendMessage at this point takes an overrides object as argument and the auto-send direction takes the selections of the workflow/tool/file before we clean up the state of the composer and sends them into the hook. The hook transmits those directly to the stream call, meaning the backend gets everything it needs for the workflow execution.

frontend/src/features/chat/components/composer/Composer.tsx
jigarbhoye04: Same update here:
the manual submission route now transmits uploadedFileData, selectedTool, selectedToolCategory, and selectedWorkflow through the overrides object, thus the stream gets the entire payload as it was earlier.

aryanranderiya:
@jigarbhoye04 

I think the responses for the conversations/:id api call could be cached because we don't want load on the server when each user loads tens of conversations on the initial page load. Fetches happen on reloading the page too. Maybe we can use react query for this?

CleanShot 2025-10-09 at 19 31 03@2x

<br />
<br />
<br />

Also, when i hopped into your branch and ran everything I got these errors, maybe it's something to look into: 

<br />
<br />

CleanShot 2025-10-09 at 19 31 38@2x

aryanranderiya:
@jigarbhoye04 any update?

aryanranderiya:
@Dhruv-Maradiya I've fixed the issue. please review once again.

PR #246, opened 2025-09-28
refactor: Enhance backend configuration and service initialization

aryanranderiya:
- Updated `.env.example` to include new environment variables for database connections, frontend URL, and authentication keys.
- Refactored `lifespan.py` to streamline service initialization and added validation for startup requirements.
- Introduced `lifespan_helpers.py` for better organization of service initialization and shutdown methods.
- Added `startup_validation.py` to check if necessary models and payment setups are completed before application startup.
- Created new scripts for payment setup and improved documentation in `setup.sh` for better clarity on running setup scripts.


backend/app/services/startup_validation.py
Dhruv-Maradiya, line 1: I think we should cache this, its not a good idea to check requirements on every single request

Dhruv-Maradiya (changes requested):

Dhruv-Maradiya (approved):

PR #251, opened 2025-10-02
feat: Minor onboarding & memory improvements

aryanranderiya:
@Dhruv-Maradiya

So a couple things:

- I removed pagination from fetching memories, previously it was fetching all and then applying pagination so it was redundant. Also, pagination wouldn't make sense with Graph Memory, because the graph is reliant on all the memories. Something we can figure out in the future we can leave it like this as of now.
- The email processor flow is working, i've tested it with the test script, but still need to test it out after connecting composio (bcz of the authConfig issue with them)
- We still haven't figured out a way to use the memories after parsing all emails right after onboarding (similiar to a gatekeeper where user can negotiate payment and they have their "aha moment")
- I've added "register_all_providers" for common lifecycle steps for both arq worker and fastapi so please check the lifespan.py and worker.py to check if the logic is proper for lazy and eager loading and if anything may be missing
- Rest are minor changes here and there


Dhruv-Maradiya:
You can use `gmail_new's`  `auth_config_id` from composio and replace it in `oauth_config` and gmail tools will work.

aryanranderiya:
@Dhruv-Maradiya should I merge? or let's wait for major improved onboarding experience

PR #362, opened 2025-12-05
refactor: Enhance integration disconnection to support self-managed integrations and auto-select all calendars by default in the calendar store.

aryanranderiya:
This pull request introduces improved handling for disconnecting integrations in the backend and enhances the user experience in the frontend calendar store. The main changes include supporting disconnection of both Composio-managed and self-managed integrations, adding robust error handling, and ensuring that all calendars are selected by default when a user connects new calendars.

### Backend: Integration Disconnection Logic

* Added support for disconnecting both Composio-managed and self-managed integrations in the `disconnect_integration` endpoint. Now, Composio-managed integrations use the Composio service for disconnection, while self-managed integrations revoke tokens internally. Unsupported integrations return a descriptive error.
* Improved error handling: specific exceptions for not found and general errors are now logged and mapped to appropriate HTTP status codes, providing clearer feedback to API consumers.
* Removed redundant error handling code at the end of the function, as error handling is now performed within each integration type branch.

### Frontend: Calendar Store Improvements

* Updated the `setCalendars` method in `calendarStore.ts` so that, on initial load or new connection, if no calendars are selected, all available calendars are automatically selected by default. This improves the user experience by ensuring users see all their calendars without extra steps.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -130,7 +130,7 @@ jobs:
           git config --global user.email "actions@github.com"
 
           echo "Running prek checks..."
-          uvx prek run --all-files --show-diff-on-failure
+          uv run prek run --all-files --show-diff-on-failure
 
           # Auto-commit fixes if changes were made
           if ! git diff --exit-code; then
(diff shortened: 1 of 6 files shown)


aryanranderiya:
Still need to comprehensively test the all calendar selection, will do after review.

Dhruv-Maradiya (approved):

PR #371, opened 2025-12-14
feat: add Voice mode

Vinitthakkar45:


backend/app/config/secrets.py
aryanranderiya, line 1: Why do we need to cache _get_infisical_client if it's an internal method not used anywhere else? this singleton pattern seems redundant

aryanranderiya: I think we can just reset this file because we need those comments too

backend/app/api/v1/endpoints/chat.py
```diff
@@ -112,3 +120,52 @@ async def save_incomplete_conversation(
...
+@router.get("/token")
+def get_token(
+    user: dict = Depends(get_current_user),
+    conversationId: Optional[str] = None,
+):
+    user_id = user.get("user_id")
+    user_email: str = user.get("email", "")
+    if not user_id or not isinstance(user_id, str):
+        return "Invalid or missing user_id"
+    room_name = f"voice_session_{user_id}_{uuid.uuid4().hex[:8]}"
```
aryanranderiya: don't truncate this with [:8]

backend/app/api/v1/middleware/agent_auth.py
aryanranderiya, line 1: move to a different file under the proper directory structure, middleware/agent_auth isn't appropriate. seems like could be a helper or util

backend/pyproject.toml
aryanranderiya, line 1: If we've created a dependency group for backend here, then you'll need to fix every reference (including but not limited to docker, gh actions, mise) where we're using uv sync or installing dependencies otherwise the backend api dependencies won't be installed.

frontend/src/features/chat/utils/voiceUtils.ts
aryanranderiya, line 1: frontend/src/features/chat/components/livekit/lib/utils.ts

don't create a utils file inside components. create under features/chat/utils and name it something like voiceUtils

.gitignore
```diff
@@ -90,5 +90,8 @@ dist/
 .pnpm-store/
 
 .vscode/settings.json
+
+uv.lock
```
aryanranderiya: why is uv.lock in .gitignore?

frontend/src/features/chat/components/interface/sections/NewChatSection.tsx
```diff
@@ -20,11 +20,13 @@ interface NewChatSectionProps {
     conversationId?: string;
   };
   showUseCases: boolean;
+  voiceModeActive: () => void;
```
aryanranderiya: voiceModeActive should be a part of the composer props, we are already destructuring it when passing the props

frontend/src/features/chat/components/interface/layouts/ChatWithMessages.tsx
```diff
@@ -48,7 +50,7 @@ export const ChatWithMessages: React.FC<ChatWithMessagesProps> = ({
       </div>
       {/* Fixed composer at bottom */}
       <div className="flex-shrink-0 pb-2">
-        <Composer {...composerProps} />
+        <Composer {...composerProps} voiceModeActive={voiceModeActive} />
```
aryanranderiya: voiceModeActive should be a part of the composer props, we are already destructuring it when passing the props

frontend/src/components/navigation/Navbar.tsx
```diff
@@ -136,7 +136,6 @@ export default function Navbar() {
                       ? "text-primary"
                       : "text-zinc-300 hover:text-zinc-100"
                   }`}
-                  as={Link}
```
aryanranderiya, line 139: ?

frontend/src/components/shared/LinkButton.tsx
aryanranderiya, line 1: Why have we omitted the as and href props? Unnecessary change, previously using the button as anchor tag and as Link is the preferable approach

backend/app/api/v1/endpoints/chat.py
Vinitthakkar45: ok

backend/app/api/v1/middleware/agent_auth.py
Vinitthakkar45: ok moved to a separate file voice_token.py

.gitignore
Vinitthakkar45: last time we discussed it should not be pushed, no worries removing it

backend/pyproject.toml
Vinitthakkar45: ya we need to specify while installing - 

uv pip install --group backend --group dev
uv pip install --group voice_agent

frontend/src/components/shared/LinkButton.tsx
Vinitthakkar45: reverting it

Vinitthakkar45: ts complier was complaining on gh actions

frontend/src/components/navigation/Navbar.tsx
Vinitthakkar45: ts complier was complaining on gh actions

frontend/src/features/chat/utils/voiceUtils.ts
Vinitthakkar45: ok

frontend/src/features/chat/components/interface/layouts/ChatWithMessages.tsx
Vinitthakkar45: ya sry my bad

frontend/src/features/chat/components/interface/sections/NewChatSection.tsx
Vinitthakkar45: ya sry my bad

backend/app/config/secrets.py
Vinitthakkar45: reverted

PR #420, opened 2026-01-06
Prevent duplicate Gmail processing by tracking processed message IDs

### LINKED ISSUE #359: fix: Memory graph prevent re-execution
When user connects the gmail we process 100s of previous email to build memory but when user disconnect and connect gmail in 5 minutes it will process them again. We need to prevent this.

## Acceptance Criteria
- [ ] Prevent re-processing of emails when reconnecting within 5 minutes.
- [ ] Ensure no duplicate memory entries are created.

Gurmeet-byte:
### Problem
When a user disconnects and reconnects Gmail, the system can reprocess previously ingested emails because it does not track individual Gmail messages. This leads to duplicate memory creation.

### Fix
This PR adds message-level deduplication by storing processed Gmail message IDs and filtering them out before processing.

### Result
- Reconnecting Gmail (including within 5 minutes) no longer reprocesses old emails  
- Only new, unseen emails are ingested  
- Duplicate memory entries are prevented

Fixes #359 

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed restriction preventing re-execution of email processing on previously synced accounts
  * Improved tracking of processed emails to prevent duplicate syncing
  * Enhanced email scan state persistence for better synchronization accuracy

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/apps/api/app/agents/memory/email_processor.py
+++ b/apps/api/app/agents/memory/email_processor.py
@@ -158,14 +158,6 @@ async def process_gmail_to_memory(user_id: str) -> Dict:
     Returns dict with processing stats.
     """
     user = await users_collection.find_one({"_id": ObjectId(user_id)})
-    if user and user.get("email_memory_processed", False):
-        logger.info(f"User {user_id} emails already processed, skipping")
-        return {
-            "total": 0,
-            "successful": 0,
-            "already_processed": True,
-            "processing_complete": True,
-        }
 
     # Extract user name for consistent memory attribution
     user_name = user.get("name") if user else None
@@ -178,7 +170,6 @@ async def process_gmail_to_memory(user_id: str) -> Dict:
 
     fetch_start_time = time.time()
     page_token = None
-    batch_count = 0
 
     # Track memory storage tasks to await completion
     email_storage_tasks = []
@@ -200,27 +191,30 @@ async def process_gmail_to_memory(user_id: str) -> Dict:
     except Exception as e:
         logger.warning(f"Failed to emit initial progress: {e}")
 
-    # Check for last scan timestamp
+    # Load Gmail scan state
     last_scan_timestamp = None
+    processed_message_ids = set()
+
     if user:
         scan_states = user.get("integration_scan_states", {})
         if isinstance(scan_states, dict):
             gmail_state = scan_states.get("gmail", {})
             if isinstance(gmail_state, dict):
                 last_scan_timestamp = gmail_state.get("last_scan_timestamp")
+                processed_message_ids = set(
+                    gmail_state.get("processed_message_ids", [])
+                )
 
     # Build query with timestamp if available
     current_query = EMAIL_QUERY
-    if last_scan_timestamp:
-        if isinstance(last_scan_timestamp, datetime):
-            timestamp_seconds = int(last_scan_timestamp.timestamp())
-            current_query = f"{EMAIL_QUERY} after:{timestamp_seconds}"
+    if last_scan_timestamp and isinstance(last_scan_timestamp, datetime):
+        timestamp_seconds = int(last_scan_timestamp.timestamp())
+        current_query = f"{EMAIL_QUERY} after:{timestamp_seconds}"
 
     try:
         while total_fetched < MAX_RESULTS:
             remaining = MAX_RESULTS - total_fetched
             batch_size = min(BATCH_SIZE, remaining)
-            batch_count += 1
 
             result = await search_messages(
                 user_id=user_id,
@@ -230,9 +224,20 @@ async def process_gmail_to_memory(user_id: str) -> Dict:
             )
 
             batch_emails = result.get("messages", [])
-
             if not batch_emails:
                 break
+
+            # Remove already processed emails
+            new_batch_emails = [
+                email for email in batch_emails
+                if email.get("id") not in processed_message_ids
+            ]
+
+            if not new_batch_emails:
+                page_token = result.get("nextPageToken")
+                if not page_token:
+                    break
+                continue
 
             # Update page token for next iteration
             page_token = result.get("nextPageToken")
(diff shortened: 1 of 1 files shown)


Gurmeet-byte:
Hello
This is my first contribution if I did any mistake please tell me

Dhruv-Maradiya:
Hey @Gurmeet-byte! 👋

First off, **thank you so much for this contribution**, it really means a lot to see community members diving into the codebase and proposing improvements! 🙏

I'm reviewing your PR and have a few questions to better understand the issue: 

**Why timestamp alone might not be sufficient? ** We're currently using `after:{timestamp}` in the Gmail query, which should filter out previously processed emails.  Did you observe cases where emails were still being reprocessed despite this? I'd love to understand the specific scenario better.

**Concern about unbounded growth:** The `processed_message_ids` set will grow indefinitely over time.  For users with large inboxes, this could become quite large. Have you considered any cleanup mechanism?

Also, I'd love to hear about your experience:

- **How was the setup process for Gaia?** Did you run into any issues or areas where documentation could be improved? 

Looking forward to hearing your thoughts! 🚀

Gurmeet-byte:
Hi @Dhruv-Maradiya, thanks for the response.

Before going deeper into the PR discussion, I wanted to highlight a blocking issue I encountered while trying to self-host GAIA locally, because it affects basic backend accessibility.

Issue: Backend never binds to a port in local self-hosting

While setting up GAIA locally, the backend currently requires all internal services to be fully operational before even starting the web server. Because of this, the API never becomes reachable.

What I did-->

1.Started infra using infra/docker/docker-compose.yml (Postgres, Mongo, Redis, RabbitMQ, Chroma all healthy)

2.Installed backend dependencies

3.Ran backend with:
uvicorn app.core.app_factory:create_app --factory --host 127.0.0.1 --port 8000

4.Backend loads settings and providers, but never opens a listening socket

5.netstat shows nothing bound to the port → /docs is unreachable

What I found-->

The backend enters unified_startup() in:

apps/api/app/core/provider_registration.py

and blocks/fails during eager service initialization:

eager_services = [
    (init_mongodb_async, "mongodb"),
    (init_reminder_service, "reminder_service"),
    (init_workflow_service, "workflow_service"),
]


Failures in these services (notably reminder/checkpointer paths) cause:

1.startup to abort

2.uvicorn to exit before binding

3.no API access, even for /docs

So currently all internal services are treated as fatal dependencies before the server can even start.

and I think this will make self hosting very difficult once this is clarify i would be able to do more and  would be able to answer your question regarding PR

Dhruv-Maradiya:
Hi @Gurmeet-byte, thanks for the detailed write-up and for taking the time to investigate this. We really appreciate the effort. 🙌

I want to clarify a few things about the local development and backend startup flow.

## Local development setup

For local development, please follow the official documentation:

- Main docs: https://docs.heygaia.io/
- Development setup: https://docs.heygaia.io/developers/development-setup#system-requirements

The documented dev setup uses a single command that:
- Starts all required infrastructure services (Postgres, Mongo, Redis, RabbitMQ, Chroma, etc.)
- Starts the backend
- Starts the frontend

This is the supported and tested path for self-hosting and avoids the startup issues you are seeing.

After correct setup, you will be able to start the API, web, and docker containers with a single command:

```bash
mise dev
````

## About running the backend with uvicorn

The command you used:

```bash
uvicorn app.core.app_factory:create_app --factory --host 127.0.0.1 --port 8000
```

is not the intended way to start the GAIA backend.

### Why this causes the issue

* `create_app` is an application factory, not the backend entrypoint.
* Running it directly initializes providers and eager services but does not start the FastAPI server lifecycle correctly.

### Correct way to start the backend

If you want to run the backend using uvicorn, please use the actual entrypoint instead (as defined in the repo), for example:

```bash
uvicorn main:app --host 127.0.0.1 --port 8000
```

This ensures the FastAPI app is created properly and the server binds to the port as expected.

## About eager service initialization

You’re right that some internal services are initialized eagerly during startup. This is intentional, as GAIA depends on these services to function correctly.

That said, your feedback around making self-hosting more forgiving and improving failure visibility is valid and appreciated. We’re open to improving this area going forward.

## Docs, feedback, and community support

We are very open to:

* Contributions that improve documentation
* Feedback on developer experience
* Suggestions that make self-hosting and onboarding easier

If you’d like to discuss issues, ask questions, or get faster feedback, feel free to reach out to us here:

* Discord: [https://discord.heygaia.io/](https://discord.heygaia.io/)
* WhatsApp: [https://whatsapp.heygaia.io/](https://whatsapp.heygaia.io/)

Once the backend is running using the documented setup, we can continue the discussion on the Gmail message deduplication PR in more detail.

Thanks again for the contribution, and looking forward to continuing the discussion 🚀

Gurmeet-byte:
Thanks for this help i was able to self host
Now lets continue our discussion on the Gmail Duplication PR 
The after:{timestamp} filter helps limit what we fetch, but it doesn’t fully prevent duplicates. In reconnect or edge cases, Gmail can still return some emails around the boundary , especially when multiple emails arrive at the same time or very close together. That’s why this PR adds message-ID checks as a second layer. Since Gmail message IDs are unique, this gives us a reliable way to skip emails we’ve already processed.

I agree that keeping all processed IDs forever isn’t ideal. Because the timestamp filter already blocks old emails, we really only need to keep IDs from a recent window where overlap can happen. We can prune IDs by age (e.g., keep only the last X days) or cap the set size so it stays bounded. I’m happy to add that cleanup logic if that approach makes sense.

Thanks for taking a look — happy to brainstorm further if needed.

Dhruv-Maradiya:
Thanks for the detailed explanation, your reasoning around the `after:{timestamp}` boundary cases makes sense, and I appreciate you thinking through the edge conditions 👍

That said, from our side, those edge cases are relatively rare in practice. Even if a few emails are occasionally reprocessed, it has not caused any functional or performance issues so far. The downstream memory layer (mem0) already performs its own deduplication, and while we do sometimes observe duplicate memories, they haven’t had any meaningful negative impact.

Additionally, users disconnecting and reconnecting Gmail itself is not a very common action. Given that, persisting and managing processed message IDs in the database feels like additional complexity and overhead compared to the value it provides. Even with cleanup logic (time-based pruning or size caps), we would still be adding extra storage, processing, and maintenance cost to handle a scenario that occurs infrequently.

From a tradeoff perspective, the current timestamp-based approach seems “good enough” for now, and we’re cautious about introducing long-term state unless it clearly solves a more significant or frequent problem.

That said, I’m open to being convinced otherwise, if there are concrete scenarios where this causes real issues (data corruption, significant user-facing problems, or noticeable performance degradation), I’d love to hear more. Let me know if I’m missing something important here.

aryanranderiya:
Closing due to inactivity @Gurmeet-byte

PR #473, opened 2026-02-13
feat: Implement Composio Google Super-Connector

### LINKED ISSUE #293: feat: Implement Composio Google Super-Connector
This issue is a follow-up to the migration of Calendar and Google Docs tools to Composio (#292 ). Once individual Google-related tools are integrated via Composio, we should aim to streamline the connection process for users.

**Objective**:
Implement a Composio Google Super-Connector that allows users to connect all Google-related tools (e.g., Calendar, Google Docs, Gmail, etc.) in a single authorization flow, rather than requiring individual connections for each tool.

**Benefits**:
- **Improved User Experience**: Users will have a much smoother and faster onboarding process for Google services.
- **Reduced Friction**: Eliminate the need for multiple authorization prompts, leading to less user fatigue.
- **Centralized Management**: Provide a single point of control for managing Google service permissions within GAIA.
- **Scalability**: Easily integrate new Google services under the same super-connector in the future.

**Action Items**:
1. Research Composio's capabilities for creating a unified connector for multiple services under a single provider (Google).
2. Design the user flow for connecting the Google Super-Connector.
3. Develop the integration logic to handle a single authorization for all relevant Google tools.
4. Update the GAIA UI to reflect the new super-connector option.
5. Thoroughly test the super-connector to ensure all integrated Google tools function correctly.
6. Document the implementation and usage of the Google Super-Connector.

miclaldogan:
## Summary

Implements a **Google Super-Connector** that allows users to connect all Google services (Calendar, Gmail, Docs, Sheets, Tasks, Meet, Maps) in a single unified flow, rather than requiring individual connections for each tool.

**Closes #293**

### Backend

| File | Change |
|---|---|
| `oauth_config.py` | Added `google` super-connector entry with `is_special=True` and `included_integrations` referencing all 7 Google services |
| `integration_connection_service.py` | Added `connect_super_connector()` — iterates unconnected child integrations and redirects to OAuth one by one; Added `disconnect_super_connector()` — disconnects all child integrations |
| `oauth_service.py` | Derives super-connector status from child statuses; Skips super-connectors from Composio batch check |
| `config.py` (endpoint) | Routes super-connector connect/disconnect through dedicated handlers |

### Frontend

| File | Change |
|---|---|
| `IntegrationsList.tsx` | Added `SuperConnectorRow` component with child integration status chips; Added "Bundles" section; Filters children from Featured/category |
| `IntegrationsCard.tsx` | Hides child integrations from sidebar, shows only super-connector |
| `toolIconConfig.ts` | Added `google` icon config using existing `google.svg` |

## How It Works

1. User clicks **"Connect All"** on the Google super-connector card
2. Backend checks which child integrations are not yet connected
3. Redirects user to OAuth for the first unconnected service (e.g., Google Calendar)
4. After OAuth callback, user returns to integrations page and clicks connect again
5. Backend redirects to the next unconnected service (e.g., Gmail)
6. Repeats until all 7 Google services are connected
7. Super-connector shows **"All Connected"** status

## Architecture

- Leverages existing `is_special` and `included_integrations` fields that were already defined in the `OAuthIntegration` model but unused
- No new database tables or API endpoints — uses existing connect/disconnect flow
- Super-connector status is derived (not stored) from child integration statuses
- Pattern is generic enough to support future super-connectors (e.g., Microsoft)

## Screenshots

The UI shows a new "Bundles" section with the Google super-connector card displaying all 7 child integrations with their connection status as small chips with green dots.

### DIFF
--- a/apps/api/app/api/v1/endpoints/integrations/config.py
+++ b/apps/api/app/api/v1/endpoints/integrations/config.py
@@ -76,12 +76,12 @@ async def disconnect_integration_endpoint(
                 resolved.platform_integration.included_integrations,
             )
         except ValueError as e:
-            raise HTTPException(status_code=400, detail=str(e))
+            raise HTTPException(status_code=400, detail=str(e)) from e
         except Exception as e:
             logger.error(f"Error disconnecting super-connector {integration_id}: {e}")
             raise HTTPException(
                 status_code=500, detail="Failed to disconnect integration"
-            )
+            ) from e
 
     try:
         return await disconnect_integration(user_id, integration_id)


Dhruv-Maradiya:
Hey @miclaldogan,

Thanks for putting this together.

Your approach of creating a new config and marking it as “special” is definitely the right direction 👍. 

We’re using Composio as our tools layer, and it already handles auth on our behalf. Composio provides a `google_super` oauth_config, which is what we should be relying on.

Using this config will bundle all required Google scopes into a single OAuth session, giving access to all Google tools at once, instead of iterating over each integration and managing scopes individually. This avoids unnecessary complexity on our side.

miclaldogan:
Hey @Dhruv-Maradiya, thanks for the feedback!

I've already updated the implementation in the latest commit to use Composio's `google_super` config. Here's what changed:

- The super-connector now uses `provider="google_super"` with its own `ComposioConfig`, so it goes through a single OAuth session instead of iterating over each child integration
- `connect_super_connector` is simplified — it just delegates to `connect_composio_integration` with the `google_super` provider
- On disconnect, it revokes the super-connector's own Composio account first, then does best-effort cleanup of any individually-connected children
- Status propagation: when the super-connector is connected, all child integrations are marked as connected too

I left the `auth_config_id` as a placeholder (`<GOOGLE_SUPER_AUTH_CONFIG_ID>`) since I don't have access to the actual Composio dashboard. Could you provide the correct `auth_config_id` for the `google_super` config so I can fill it in?

Also addressed all CodeRabbit feedback in the same commit (error handling, disconnect failure tracking, search/count bugs, accessibility).

Dhruv-Maradiya:
Hey @miclaldogan  

While testing your changes locally I realized one thing that Composio Super is itself a integration it doesn't expose child integrations like Gmail, Calendar, Maps, Meet, Docs etc.
And our codebase heavily rely on these integrations.
So at this point we can't add this feature. 

Thank you for your contribution,

PR #599, opened 2026-04-05
feat(analytics): add PostHog analytics to bots

aryanranderiya:
## Summary

- Adds `Analytics` class to `@gaia/shared` wrapping `posthog-node` for server-side event tracking in bots
- Instruments `BaseBotAdapter` (`dispatchCommand`, `shutdown`, auth flows) and `handleStreamingChat` with `bot:*` events; all four bot adapters pass `this.analytics` through
- CLI ships telemetry-free so self-hosters get a clean binary (CLI instrumentation was added then pulled before merge)

## Event taxonomy

`bot:message_received`, `bot:command_executed`, `bot:chat_started`, `bot:chat_completed`, `bot:auth_initiated`, `bot:error`

`message_received` fires for `/gaia`, DMs, and mentions so the top of the funnel includes chat interactions, not just slash commands.

## Privacy hardening

- `disableGeoip: true` on the PostHog Node client
- No `identify()` call — no username or display_name PII sent
- Errors ship `error_type` (error.name) instead of raw `error_message`
- `posthog-node` pinned to exact `5.28.11`

## Bug fixes along the way

- `bot:chat_completed` no longer fires on auth failures (wraps `onAuthError` to set `hadError`)
- Redis CI configured with 32 databases to support xdist worker isolation
- `vulture_whitelist.py` ruff F821 / mypy name-defined suppressions
- Duplicate `esbuild` entries removed from `pnpm-lock.yaml`

## Test plan

- [x] All 292 bot adapter tests pass (`nx test bots-e2e`)
- [x] Type-check passes on all four bots and shared lib
- [x] Biome lint passes on all affected projects
- [x] Analytics is a silent no-op when `POSTHOG_API_KEY` is unset (safe for local dev)
- [ ] Set `POSTHOG_API_KEY` in Infisical (`development` + `production`) to activate in deployed bots


aryanranderiya:
@coderabbitai review

PR #617, opened 2026-04-07
fix: resolve streaming race condition and missing UI messages

aryanranderiya:
## Summary
Fix event-loop race condition during streaming and resolve UI missing messages.

## Context
When sending multiple messages in an existing conversation, the frontend UI would fail to render subsequent messages despite data arriving in the Network tab. The frontend failed to establish a connection with the backend because the background task created and published the initial chunk to Redis before the HTTP Response handler had fully subscribed to the channel. This caused the frontend to silently skip updating its state.

## Changes
- Replaced hacky `asyncio.sleep` with a robust `asyncio.Event()` synchronization between the background streaming task and the HTTP handler.
- The `start_event` guarantees the backend only publishes chunks *after* the HTTP connection is fully subscribed to the Redis channel, ensuring zero dropped chunks regardless of CPU load or network latency.
- Added strict conversation branch handling in `apps/web/src/features/chat/hooks/useChatStream.ts` to correctly track existing chats instead of miscategorizing them as brand-new conversations.
- Fixed a React UI crash in `WeatherCard.tsx` that occurred during progressive streaming of incomplete weather objects.

## Testing
- Sent multiple test messages on the frontend via Chrome DevTools MCP, verifying that subsequent messages correctly update the chat UI without hard reloads.
- Verified that weather component renders successfully and handles partial streaming data gracefully.

### DIFF
--- a/.mcp.json
+++ b/.mcp.json
@@ -4,6 +4,11 @@
       "command": "dagger",
       "args": ["-s", "mcp"],
       "env": {}
+    },
+    "chrome-devtools": {
+      "command": "npx",
+      "args": ["-y", "chrome-devtools-mcp@latest", "--autoConnect"],
+      "env": {}
     }
   }
 }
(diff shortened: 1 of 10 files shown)


Dhruv-Maradiya (approved):

aryanranderiya:
@coderabbitai implications of these event related changes?

PR #644, opened 2026-04-17
fix(api): harden workflow timezone resolution + lifecycle wide events

aryanranderiya:
## Summary
- Three-tier timezone fallback (`user.timezone` → `x-timezone` header → UTC) in `get_user_timezone_from_preferences`, with fire-and-forget Mongo backfill so users who skipped onboarding get populated on next request.
- Emits `timezone_source` on every resolution — the silent-UTC drift that caused scheduled workflows to fire hours off for non-UTC users is now visible in wide events instead of invisible.
- Instruments the full workflow lifecycle (create → register → webhook → enqueue → execute) with structured wide-event fields so a single trace explains timing end-to-end.
- Warns when webhook lag or execution drift exceeds 5 minutes — future timing reports point at the faulty hop immediately.

## Wide-event fields added

| Stage | File | Fields |
|---|---|---|
| Create | `workflow/service.py` | `trigger_type`, `trigger_name`, `cron_expression`, `trigger_timezone`, `next_run_utc` |
| Register | `triggers/handlers/calendar.py` | `composio_slug`, `calendar_count`, `minutes_before_start`, `composio_trigger_ids` |
| Webhook | `triggers/base.py` | `event_start_time_utc`, `seconds_until_event`, `countdown_window_minutes`, `webhook_lag_seconds` |
| Enqueue | `queue_service.py`, `scheduler_service.py` | `scheduled_at_utc`, `defer_seconds`, `scheduled_at_was_naive`, `arq_job_id` |
| Execute | `workers/tasks/workflow_tasks.py` | `actual_fire_utc`, `drift_from_scheduled_seconds` |

## Test plan
- [ ] Sign in as a user whose Mongo doc has no `timezone` field and a valid `x-timezone` header — confirm the header value is persisted to Mongo on next request and wide event shows `timezone_source=x_timezone_header`.
- [ ] Confirm existing users with `user.timezone` set continue to see `timezone_source=user_profile`.
- [ ] Create a scheduled workflow with a cron in a non-UTC zone and verify `next_run_utc` matches the expected local time converted to UTC.
- [ ] Trigger a `calendar_event_starting_soon` webhook and verify `webhook_lag_seconds` is close to 0; delayed webhooks should log the drift warning.
- [ ] Run `nx type-check api` and `nx lint api` locally.

### DIFF
--- a/apps/api/app/api/v1/dependencies/oauth_dependencies.py
+++ b/apps/api/app/api/v1/dependencies/oauth_dependencies.py
@@ -1,8 +1,44 @@
-from datetime import datetime
-from zoneinfo import ZoneInfo
+import asyncio
+from datetime import datetime, timezone as dt_timezone
+from typing import Any
+from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
 
+from bson import ObjectId
 from fastapi import Depends, Header, HTTPException, Request, WebSocket, status
 from shared.py.wide_events import log
+
+from app.db.mongodb.collections import users_collection
+
+_TIMEZONE_BACKFILL_TASKS: set[asyncio.Task[Any]] = set()
+
+
+def _is_valid_iana_or_offset(tz: str) -> bool:
+    """Cheap validation: accept IANA names or ±HH:MM offsets; reject garbage."""
+    if not tz:
+        return False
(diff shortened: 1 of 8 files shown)


aryanranderiya:
@coderabbitai review

PR #646, opened 2026-04-18
feat(retrieval): pass intent state in LangGraph and invalidate service cache after user connects, add lazy hydration of tools

sankalpaacharya:
## Cache Validation

When a user first connects to any services (GitHub, Twitter, Gmail) their connection status gets stored in 
the Redis cache and when we retrieve the tool from ChromaDB we do a search from the namespace. Suppose if 
the user query is `how many followers do i have on github` the namespace for ChromaDB would be `github` 
and the rest of the services tool information will be queried but the Redis cache validation was not 
working properly. Now when internally we search which services the user is connected to it shows false 
connected apps and the namespace.

Below is the log I did for the retrieval system and even though GitHub is connected there is no namespace 
so when we retrieve a tool from ChromaDB there are no GitHub tools, you can see that in the image except 
the auto bind ones I guess cause it's a semantic search on vector db it should return tools with lower 
similarity no matter how bad my query was but the query was clear and tools are empty. So I removed the 
cache from Redis and it worked, it started showing. So I added a Redis cache validation which validates 
the cache at the time when the user actually gets the service.

```
timestamp: 2026-04-18T12:48:38.120684+00:00
mode: discovery
tool_space: general
user_id: 69e269059871f1ffc0b8bbc0
query: "github followers"
stats:
  chroma_hits: 40
  public_hits: 0
  candidates: 40
  tools_selected: 29
chroma_preview:
  - ('general',)::list_workflows
  - ('general',)::suggest_integrations
  - ('general',)::get_notification_count
  - ('general',)::search_triggers
  - ('general',)::get_notifications
  - ('general',)::gather_context
  - ('general',)::search_notifications
  - ('general',)::query_file
  - ('general',)::mark_notifications_read
  - ('general',)::vfs_cmd
namespaces:
  - general
  - goals
  - reminders
  - skills
  - subagents
  - todos
```

Before

image

After
image

## Query Drift

image

We are only using tools which are registered from the tool registry, so I tried to be explicit here 
`GITHUB_GET_USER` which exists in Composio GitHub tools but not in our registry, it was expected that on 
the next query retrieval it removed that tool name. But the problem is sometimes it starts to drift from 
the original intent to solve the problem leading to choosing the wrong tools. If my original query was 
`how many followers I have on GitHub` after 2-3 deep tool retrievals it starts searching my repo. For this 
I have added an additional variable on the LangGraph node state which is the original query as an intent 
so that it remembers what our goal is.

## Assuming Wrong Usernames

You can see the output `the user checking server` that's not my GitHub username, it's the username I have 
in Gaia. Users can have different names as per different connected services, so I made a little tweak on 
the prompt and on `handoff_tools.py` to explicitly mention the service name alongside the Gaia username.

image

## Tool Budget (to be added)

Right now the agent runs in a loop to solve a problem even though it knows such a tool doesn't exist, it 
should stop trying if it doesn't have enough tools to fulfill that request. We can do this by adding a 
tool budget system.


aryanranderiya:
@sankalpaacharya Ready to merge?

sankalpaacharya:
> @sankalpaacharya Ready to merge?

yes sir!

apps/api/app/agents/core/subagents/base_subagent.py
```diff
@@ -108,6 +108,13 @@ async def create_provider_subagent(
         scoped_tool_dict[fetch_webpages.name] = fetch_webpages
         scoped_tool_dict[deep_research.name] = deep_research
 
+        @tool(description="Finish the task and return the final result to the parent.")
```
Dhruv-Maradiya: This is not the place to create a tool can be moved to tools folder

apps/api/app/agents/core/subagents/base_subagent.py
```diff
@@ -164,6 +171,8 @@ async def create_provider_subagent(
             log.info(
                 f"Auto-binding {len(valid_auto_bind)} tools for {provider}: {valid_auto_bind}"
             )
+        print(f"DEBUG valid_auto_bind for {provider}: {valid_auto_bind}",flush=True)
```
Dhruv-Maradiya: Let's remove this debug logs

apps/api/app/agents/middleware/subagent.py
```diff
@@ -233,6 +233,11 @@ async def _execute_subagent(
             config=config,
             inherited_tool_names=inherited_tool_names,
         )
+        @tool(description="Finish the task and return the final result to the parent.")
```
Dhruv-Maradiya: Again same pattern can reuse this

apps/api/app/agents/middleware/subagent.py
```diff
@@ -256,6 +261,14 @@ async def _execute_subagent(
             if not response.tool_calls:
                 return str(response.content) if response.content else "Task completed."
 
+            for tc in response.tool_calls:
```
Dhruv-Maradiya, line 257: This logic should live in create_agent.py

apps/api/app/agents/tools/core/retrieval.py
```diff
@@ -35,6 +38,104 @@
 from langgraph.store.base import BaseStore, SearchItem
 
 WEBPAGE_TOOLS = [web_search_tool.name, fetch_webpages.name, deep_research.name]
+_RETRIEVAL_LOG_PATH = Path("logs/retrieval-debug.log")
```
Dhruv-Maradiya: Debug logs?

apps/api/app/agents/tools/core/retrieval.py
```diff
@@ -35,6 +38,104 @@
...
+    for integration in OAUTH_INTEGRATIONS:
+        if (
+            integration.subagent_config
+            and integration.subagent_config.tool_space == tool_space
+        ):
+            return integration
+    return None
+
+
+async def _maybe_hydrate_missing_tools(
```
Dhruv-Maradiya, line 56: Can you please explain me what this function is doing?

Dhruv-Maradiya (changes requested):

apps/api/app/agents/tools/core/retrieval.py
sankalpaacharya: We now use ChromaDB results as “hints” to load missing Composio tools on demand. During `retrieve_tools`, if Chroma returns tool names in the current `tool_space` that are **not present in the in‑memory registry**, we fetch those tools by name from Composio, register them into the correct category, and re‑index them. This closes the gap where Chroma could discover tools but the registry couldn’t bind them, which previously caused repeated retrieval attempts and tool‑search loops.

The registry is the only source of executable tools, but we were only registering a **small allow‑list** (`specific_tools`) for integrations like GitHub. That meant Chroma could return valid tools that weren’t actually loaded, so discovery produced candidates that were immediately filtered out (`candidates=0`). The agent kept re‑retrieving and drifting into general/web search even after a relevant tool existed.

Hydration exists to bridge that gap: if Chroma returns GITHUB_LIST_* but the registry doesn’t have it loaded yet, hydration pulls the tool by name (from Composio) and registers it on the fly so it can be bound(like we were having for cache) just in case if that ever happens 

Hydration is mainly for edge cases:
- Cold start on a fresh worker
- First request after a new integration is connected

this is optional just a nice to have if you want i can get rid of it no issue

apps/api/app/agents/core/subagents/base_subagent.py
sankalpaacharya: done

Dhruv-Maradiya:
Ok

PR #668, opened 2026-05-03
fix(api): invalidate tool_namespaces cache on integration status change

Dhruv-Maradiya:
## Summary

PR #646 worked around a stale `tool_namespaces:{user_id}` cache by adding ad-hoc invalidation in `oauth_service.py`, a lazy-hydration band-aid in `retrieve_tools`, and left a debug `print` in `chroma_tools_store.py`. The actual root cause was that `update_user_integration_status` only invalidated `tools:user:*`, not `tool_namespaces:{user_id}` — so any state change going through it (OAuth connect, MCP token refresh, disconnect) left retrieval blind to the user's new integrations for up to 24h.

This PR fixes it at the source by adding `tool_namespaces:{user_id}` to that function's `@CacheInvalidator`, then removes the workarounds. It also hardens `retrieve_tools` without weakening the security boundary that protects user-owned MCPs:

- `_get_user_context` seeds `tool_space` into `user_namespaces` only for hardcoded platform integrations (`OAUTH_INTEGRATIONS`); custom MCP namespaces stay gated by the cache lookup so one user cannot search another user's tool index.
- Cache contents are unioned (not assigned) into `user_namespaces` so a stale cache cannot drop the platform seed.
- Binding mode now logs a warning when it silently drops unknown tool names, and search refusals for non-owned namespaces are logged — both previously silent failures.

## Test plan

- [ ] Wipe `tool_namespaces:<user_id>` in Redis, send a chat message (populates cache without GitHub), connect GitHub, send "list my GitHub repos" — should succeed without restarting or waiting for TTL.
- [ ] Repeat with a poisoned cache (`SET tool_namespaces:<user_id> '["general","subagents"]'`) and confirm the platform seed still routes the GitHub search.
- [ ] Confirm a custom MCP belonging to user A is not searchable from user B (look for `retrieve_tools refused search: tool_space not in user_namespaces`).
- [ ] `nx lint api` and `nx type-check api`.

### DIFF
--- a/apps/api/app/agents/core/subagents/base_subagent.py
+++ b/apps/api/app/agents/core/subagents/base_subagent.py
@@ -31,6 +31,7 @@ from app.agents.tools.core.tool_runtime_config import (
 )
 from app.agents.tools.memory_tools import search_memory
 from app.agents.tools.finish_task_tool import finish_task
+from app.constants.general import FINISH_TASK_NAME
 from app.agents.tools.todo_tools import create_todo_pre_model_hook, create_todo_tools
 from app.agents.tools.vfs_tools import vfs_cmd, vfs_read
 from shared.py.wide_events import log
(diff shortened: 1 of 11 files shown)


Dhruv-Maradiya:
@coderabbitai please review

PR #724, opened 2026-05-28
fix(api): re-arm recurring workflows + persist schedule times as dates

aryanranderiya:
## Problem

Reported by a user: *"workflows are not executing even though they are scheduled daily. 3 days no notifications."*

Investigated via prod Loki + Mongo. Recurring workflows fire **once** after creation and then go silent. **19 users** affected; oldest stuck workflow hadn't run in ~6 months. Reminders share the same latent bug but mostly escape it because they're one-time.

## Root causes (both in the shared scheduler)

1. **Recurrence was never re-armed.** The workflow ARQ executor (`execute_workflow_by_id`) runs a single occurrence and never schedules the next one. Only the *reminder* path re-arms (via `process_task_execution` → `handle_recurring_task`); workflows bypass it. Result: `occurrence_count=0` on every executed recurring workflow; zero `"Rescheduled recurring task"` log lines in 8 days.
2. **`scheduled_at` persisted as a string.** `BaseScheduledTask`'s datetime `field_serializer` ran in all modes, and create used `model_dump(mode="json")` → ISO strings in Mongo. The recovery scan (`scheduled_at: {"$lte": now}`) **cannot match a string**, so 1838/1843 active workflows were invisible to it.

## Fix (unify, root-level — no patches)

- **Re-arm via the shared core.** `_handle_recurring_task` → public `handle_recurring_task`, called by both the reminder path and the workflow executor. The executor re-arms **only for scheduler-originated fires** (`trigger_type=schedule`) so a manual "run now" never shifts the schedule. Re-arms on success *and* failure so a transient error doesn't kill the series.
- **Persist datetimes natively.** Scope `serialize_datetime` to `when_used="json"` (API responses still emit ISO) and create workflows in python mode → BSON dates the scan can match.
- **Advance from now**, not a stale `scheduled_at` — a dormant workflow resumes at its next future run instead of replaying every missed run (no catch-up storm).
- **Deterministic ARQ `_job_id`** per (task, fire-time) — prevents duplicate-fire stacking across scans/edits.

## Migration

`apps/api/scripts/migrate_workflow_schedule_types.py` — idempotent, dry-run by default (`--apply` to write). Converts string → datetime and re-arms active recurring workflows.

```
cd apps/api && uv run python scripts/migrate_workflow_schedule_types.py          # dry run
cd apps/api && uv run python scripts/migrate_workflow_schedule_types.py --apply  # apply
```

## Verification

- `nx type-check api` ✅  `nx lint api` ✅
- `tests/unit/services/test_scheduler_service.py` — 29 passed (updated the rename + new `_job_id` assertion).
- Full unit run: the 10 pre-existing failures + 1 error are unchanged vs clean `develop` (slug/integration-trigger/oauth tests, unrelated to this change).
- End-to-end VM test of the full workflow lifecycle is being run separately.

Also includes an engineering-discipline note in `CLAUDE.md` (requested): always fix at the root / unify divergent paths rather than patching.

Separately tracked (not this PR): RabbitMQ `gaia` broker user missing (live WebSocket toasts) and the silent `_publish_to_rabbitmq` swallow.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -5,6 +5,7 @@ GAIA is a proactive personal AI assistant — full-stack Nx monorepo with a Next
 ## Engineering Discipline
 
 - Use best practices and write clean, idiomatic code every time. No shortcuts, no half-measures.
+- **Never ship workarounds, patches, or band-aid fixes. Always choose the cleanest, most correct approach — every single time.** When you find a bug, fix it at the root, not at the symptom. If two code paths diverge and one is broken, unify them rather than patching the broken one in place. Surgical-but-duplicative is not "safer" — it is how the bug got there. Surface the tradeoff, then take the clean path.
 - Do not override or work around the architecture. Never disable lint rules, add blanket `# noqa` / `// biome-ignore` / `# type: ignore`, or bypass CI to force something through. Linting, type-checking, and CI are guardrails that exist for a reason. Fix the cause, not the symptom.
 - Match the conventions of the surrounding code. Prefer the existing pattern over inventing a new one.
 
--- a/apps/api/app/models/reminder_models.py
+++ b/apps/api/app/models/reminder_models.py
@@ -108,9 +108,10 @@ class CreateReminderRequest(BaseModel):
                 raise ValueError("stop_after must be in the future")
         return v
 
-    @field_serializer("scheduled_at", "stop_after", "base_time")
+    @field_serializer("scheduled_at", "stop_after", "base_time", when_used="json")
     def serialize_datetime(self, value: datetime | None) -> str | None:
-        """Serialize datetime fields to ISO format strings."""
+        """ISO strings for JSON only; python mode keeps native datetimes so a
+        model_dump() used to build a Mongo document never persists a string date."""
         if value is not None:
             return value.isoformat()
         return None
@@ -285,9 +286,11 @@ class UpdateReminderRequest(BaseModel):
                 raise ValueError("stop_after must be in the future")
         return v
 
-    @field_serializer("scheduled_at", "stop_after")
+    @field_serializer("scheduled_at", "stop_after", when_used="json")
     def serialize_datetime(self, value: datetime | None) -> str | None:
-        """Serialize datetime fields to ISO format strings."""
+        """ISO strings for JSON only; python mode (the Mongo `$set` update path
+        in update_reminder) keeps native datetimes so the persisted scheduled_at
+        stays a BSON date the `$lte` recovery scan can match."""
         if value is not None:
             return value.isoformat()
         return None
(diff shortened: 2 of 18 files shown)


aryanranderiya:
## Root cause (verified on prod via Loki + Mongo)

Recurring workflows fired **once** after creation, then never again. Two shared defects in the scheduler:

1. **Recurrence was never re-armed.** The workflow ARQ executor (`execute_workflow_by_id`) ran a single occurrence and never scheduled the next one. Only the *reminder* path re-armed (via `process_task_execution → handle_recurring_task`); workflows bypassed it. Evidence: every executed recurring workflow had `occurrence_count = 0`, and `"Rescheduled recurring task"` appeared **zero times** in 8 days of prod logs.
2. **`scheduled_at` was persisted as a string.** `BaseScheduledTask`'s datetime serializer ran in all modes and create used `model_dump(mode="json")`, so Mongo stored ISO strings. The recovery scan `scheduled_at: {"$lte": now}` **cannot match a string**, so 1838/1843 active workflows were invisible to it. Reminders had the same latent bug but mostly escape it (they're one-time, served by the create-time deferred job).

These compounded: a workflow had two chances to keep going (re-arm, or the scan safety-net) and both failed.

## How it was tested

**1. Unit** — `tests/unit/services/test_scheduler_service.py` 29/29 pass; full suite shows the same pre-existing failures as clean `develop` (zero new).

**2. Integration e2e against real Mongo + Redis (OrbStack), 19/19** — only the external LLM/user-lookup mocked; everything else real:
- create → `scheduled_at`/`next_run` persist as **datetime**, `created_at` stays string, `$lte` scan finds it
- scheduled fire re-arms: `occurrence_count` 0→1→2, `scheduled_at` advances to a future datetime, stays `scheduled` (success **and** failure paths)
- manual "run now" does **not** advance the schedule
- deterministic `_job_id` dedupes duplicate enqueues
- migration repairs a legacy string-typed stuck workflow (reproduced the bug: scan misses it → migration → scan finds it)

**3. Live ARQ worker (the real user flow)** — started the actual `arq` worker against the OrbStack stack with a real every-minute workflow. No direct calls; the worker fired it on its own deferred timer:
```
02:28:36  execute_workflow_by_id('wf_147f653e1a74', {'trigger_type': 'schedule'})   # fired by timer
02:28:43  Rescheduled recurring task wf_147f653e1a74 for 2026-05-28 20:59:00         # re-armed
          'Workflow wf_147f653e1a74 executed successfully with 2 messages'           # real LLM ran
occ 0 (20:51) → 1 (20:59) → 2 (21:00)                                                # fired again on its own
```
The same run also proved the scan fix: the worker's startup scan found and executed migration-repaired (datetime-typed, previously-invisible) workflows, delivering "Workflow Completed" notifications.

**4. Migration dry-run on prod (read-only)** — 1883 type-fixes, 53 re-arms to correct future times, 1850 skipped.

After merge, run the migration with `--apply` to repair the existing 1838 stuck workflows.

aryanranderiya:
Superseded by #736, which fully includes these commits.

PR #735, opened 2026-06-03
chore(infra): add FUSE capabilities to prod stack for JuiceFS/sandbox

Dhruv-Maradiya:
## Summary

- Adds `CAP_SYS_ADMIN` + `apparmor:unconfined` + `/dev/fuse` bind-mount to `gaia-backend` and `arq_worker` in the production Swarm compose, enabling JuiceFS FUSE mounts inside both containers.
- Uses the volume bind-mount workaround for `/dev/fuse` instead of `privileged: true`, since Docker Swarm does not support the `devices:` directive — this grants only the minimum surface needed for FUSE without exposing all host devices or capabilities.
- Adds a shared `juicefs_cache` named volume (mounted at `/var/cache/juicefs`) so both services share the local read cache, reducing redundant R2 fetches.

## Test plan

- [ ] Verify `/dev/fuse` exists and FUSE is available on the VPS (`ls -la /dev/fuse`)
- [ ] Run `juicefs format` against CockroachDB + R2 (one-time, before first deploy)
- [ ] Confirm all Infisical prod secrets are set: `E2B_API_KEY`, `E2B_TEMPLATE_ID`, `R2_*`, `JUICEFS_META_URL_TEMPLATE`, `JUICEFS_NUM_SHARDS`, `ARTIFACT_DETECTION_MODE`
- [ ] Deploy via CI/CD and check API logs for `JuiceFS mount ready` on startup

### DIFF
--- a/infra/docker/docker-compose.prod.yml
+++ b/infra/docker/docker-compose.prod.yml
@@ -35,10 +35,6 @@ services:
     image: ghcr.io/theexperiencecompany/gaia:latest
     working_dir: /app/apps/api
     command: python -m uvicorn app.main:app --host 0.0.0.0 --port 80 --no-access-log --workers 1 --timeout-keep-alive 65 --timeout-graceful-shutdown 30 --h11-max-incomplete-event-size 16777216
-    # JuiceFS requires CAP_SYS_ADMIN to manage FUSE mounts and apparmor:unconfined
-    # to allow the fuse kernel module. /dev/fuse is bind-mounted as a volume because
-    # Docker Swarm does not support the `devices:` directive — this is the standard
-    # Swarm workaround for FUSE access without full privileged mode.
     cap_add:
       - SYS_ADMIN
     security_opt:
@@ -115,8 +111,6 @@ services:
     image: ghcr.io/theexperiencecompany/gaia:latest
     working_dir: /app/apps/api
     command: python -m arq app.worker.WorkerSettings
-    # Same FUSE requirements as gaia-backend — worker runs sandbox eviction tasks
-    # and shares the JuiceFS bootstrap on startup.
     cap_add:
       - SYS_ADMIN
     security_opt:
--- a/infra/docker/docker-compose.prod.yml
+++ b/infra/docker/docker-compose.prod.yml
@@ -35,6 +35,17 @@ services:
     image: ghcr.io/theexperiencecompany/gaia:latest
     working_dir: /app/apps/api
     command: python -m uvicorn app.main:app --host 0.0.0.0 --port 80 --no-access-log --workers 1 --timeout-keep-alive 65 --timeout-graceful-shutdown 30 --h11-max-incomplete-event-size 16777216
+    # JuiceFS requires CAP_SYS_ADMIN to manage FUSE mounts and apparmor:unconfined
+    # to allow the fuse kernel module. /dev/fuse is bind-mounted as a volume because
+    # Docker Swarm does not support the `devices:` directive — this is the standard
+    # Swarm workaround for FUSE access without full privileged mode.
+    cap_add:
+      - SYS_ADMIN
+    security_opt:
+      - apparmor:unconfined
+    volumes:
+      - /dev/fuse:/dev/fuse
+      - juicefs_cache:/var/cache/juicefs
     environment:
       PYTHONUNBUFFERED: "1"
       FORCE_COLOR: "0"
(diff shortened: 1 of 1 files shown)


aryanranderiya:
## Why `security_opt: apparmor:unconfined` was removed + how FUSE actually gets enabled on prod

Verified on the prod Swarm node (`gaia-backend`, Docker 29.4.0, single active node):

- **Swarm silently drops `security_opt`.** `docker stack deploy` ignores it (along with `devices`/`privileged`), so `apparmor:unconfined` never reached the containers — it only *implied* a protection that wasn't in effect. `cap_add: SYS_ADMIN` and the `/dev/fuse` bind-mount **are** honored on Docker 29, so those stay.
- **`docker-default` blocks the FUSE mount.** Proven with a throwaway container test on the node:
  - under `docker-default`: `mount(fuse)` → `EACCES`
  - with apparmor relaxed: mounts fine
- The non-root `appuser` is fine — JuiceFS mounts via setuid `fusermount3` (the `fuse3` package), not a direct `mount()` syscall.

Because Swarm can't carry a per-service profile, AppArmor must be relaxed **at the node level**. The secure way is a custom `docker-default` that is the stock moby template with the single `deny mount,` line replaced by FUSE-only allows — every other restriction is preserved, all non-FUSE mounts stay denied.

### Install on every node that runs these services (currently just `gaia-backend`)

```bash
sudo tee /etc/apparmor.d/docker-default > /dev/null <<'PROFILE'
#include <tunables/global>

# GAIA customised docker-default: stock moby template with the blanket
# `deny mount,` replaced by FUSE-only allows so JuiceFS can mount inside
# containers. Every other fstype stays denied-by-default. Swarm ignores
# per-service security_opt, so relaxing node-wide docker-default is the
# only way to permit the mount under `docker stack deploy`.
profile docker-default flags=(attach_disconnected,mediate_deleted) {
  #include <abstractions/base>

  network,
  capability,
  file,
  umount,
  signal (receive) peer=unconfined,
  signal (receive) peer=runc,
  signal (receive) peer=crun,
  signal (receive) peer=unconfined,
  signal (send,receive) peer=docker-default,

  deny @{PROC}/* w,
  deny @{PROC}/{[^1-9],[^1-9][^0-9],[^1-9s][^0-9y][^0-9s],[^1-9][^0-9][^0-9][^0-9/]*}/** w,
  deny @{PROC}/sys/[^k]** w,
  deny @{PROC}/sys/kernel/{?,??,[^s][^h][^m]**} w,
  deny @{PROC}/sysrq-trigger rwklx,
  deny @{PROC}/kcore rwklx,

  # --- GAIA JuiceFS: allow FUSE mounts only; all other mount types stay denied ---
  mount fstype=fuse,
  mount fstype=fuse.*,
  # ------------------------------------------------------------------------------

  deny /sys/[^f]*/** wklx,
  deny /sys/f[^s]*/** wklx,
  deny /sys/fs/[^c]*/** wklx,
  deny /sys/fs/c[^g]*/** wklx,
  deny /sys/fs/cg[^r]*/** wklx,
  deny /sys/firmware/** rwklx,
  deny /sys/devices/virtual/powercap/** rwklx,
  deny /sys/kernel/security/** rwklx,

  ptrace (trace,read,tracedby,readby) peer=docker-default,
}
PROFILE

sudo apparmor_parser -r -W /etc/apparmor.d/docker-default
sudo aa-status | grep docker-default   # confirm loaded, enforce mode
```

Robust across reboots (`apparmor.service` auto-loads `/etc/apparmor.d/*`) and docker restarts (moby skips its built-in template if a `docker-default` profile is already loaded). No `docker restart` needed.

### Still required for this to work end-to-end

1. **Install the profile on every node that can schedule `gaia-backend` / `arq_worker`.** Today that's only `gaia-backend` (`swarm-worker` is Down/Drain). Pin placement to a labeled node so a task never lands on an unprepared one.
2. **Redeploy the image** — the running `gaia:latest` predates the Dockerfile that installs the `juicefs` binary, so it must be rebuilt/redeployed.
3. **Meta store:** dev uses Postgres (`gaia_juicefs_{shard}`), not CockroachDB as the checklist says — reconcile.

### Verify after deploy

```bash
docker service logs gaia-prod_gaia-backend | grep -i juicefs   # expect "[juicefs] mounted"
```

> Note: an alternative cleaner architecture is to mount JuiceFS once on the host via a systemd unit and bind-propagate `/mnt/jfs` into the containers (`propagation: rslave`), dropping `cap_add`/`/dev/fuse` entirely and leaving `docker-default` untouched. `bootstrap.py` already no-ops when `/mnt/jfs` is pre-mounted, so it needs no app change. Tradeoff: R2/meta secrets must also live on the host.

PR #741, opened 2026-06-08
fix(api,web): return 403 for disconnected integrations so logged-in users aren't shown the login modal

aryanranderiya:
## Problem

Visiting **/dashboard** while logged in showed a "please log in" modal, even though the session was valid (chat and other pages worked fine).

## Root cause

The dashboard fetches `/calendar/events` (chat never does). When the Google Calendar integration has no active connection, the Composio proxy raised **`401 Unauthorized`** (`"No active GOOGLECALENDAR connection"`). The web axios interceptor treats **every** `401` as session expiry → opens the login modal.

`401` means *"not authenticated."* A missing integration connection is *"authenticated, but a precondition isn't met"* → that's a **`403`**. The frontend already renders `403 + {type:"integration"}` as an actionable toast.

## Changes

**Backend**
- `proxy_client`: integration-not-connected and provider-token-rejected now return **403** tagged `INTEGRATION_NOT_CONNECTED` (was 401).
- `get_current_user`: genuine auth 401s tagged `NOT_AUTHENTICATED`.
- `calendar_service` / `require_integration`: emit the structured `{type:"integration", ...}` detail the web client already understands.

**Frontend**
- Interceptor opens the login modal **only** for `401 + NOT_AUTHENTICATED`; integration 403s show a single deduped "Reconnect" toast.

## Verification
- `nx lint api`, `nx type-check api`, `nx type-check web` all pass.

### DIFF
--- a/apps/api/app/api/v1/dependencies/google_scope_dependencies.py
+++ b/apps/api/app/api/v1/dependencies/google_scope_dependencies.py
@@ -76,6 +76,7 @@ def require_integration(integration_short_name: str):
                 detail = {
                     "type": "integration",
                     "error_code": INTEGRATION_NOT_CONNECTED,
+                    "toolkit": integration_short_name,
                     "message": f"Missing connection: {integration_config.name}. Please connect integrations in settings.",
                 }
                 raise HTTPException(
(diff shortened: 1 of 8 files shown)


aryanranderiya:
Both addressed in c812ac38f:
- `INTEGRATION_NOT_CONNECTED` added to `apps/web/src/lib/api/errorCodes.ts` (mirrors the backend contract).
- `toolkit` (the integration short name) now included in `require_integration`'s 403 detail, so the web reconnect toast dedupes per-integration instead of falling back to the generic key.

@coderabbitai resolve

PR #742, opened 2026-06-09
feat(api): connect tool ID-only + bot-aware integration connect URL

Dhruv-Maradiya:
Tightens the `connect_integration` tool to accept exact integration IDs only (drops name/short_name fuzzy matching), with normalization moved before the loop. Removes the dead fallback URL construction in `build_integration_connection_message` (the `connect_url` is always provided by callers) along with the now-unused `INTEGRATIONS_CONNECT_PATH` constant and `settings` import. Adds surface-aware messaging: UI clients receive the connect URL silently with a no-render instruction; bot/text-only clients get an explicit instruction to pass the URL verbatim through to comms. The text-only platform addendum in `agent_template` now tells comms to paste connect URLs directly when no UI connect button exists.


Dhruv-Maradiya (commented):
Fixed — using  for an order-preserving dedup with empty-string filtering in the same line as normalization.

PR #751, opened 2026-06-12
perf(workspace): event-driven workspace materialization + chat title fix

Dhruv-Maradiya:
Moves per-user workspace (JuiceFS VFS) materialization off the chat hot path and onto the events that actually change it — user registration, integration connect/disconnect, conversation creation, and app startup — instead of running the full bootstrap on every chat turn (todos/tasks already self-heal via their existing mutation schedulers). Also fixes the auto-generated conversation title reverting to "New Chat" after a refresh: the title-update task was racing the conversation-row insert, so the `$set` matched zero rows and the title only ever lived in the SSE stream.

Adds a developer CLI (`app.scripts.sync_user_workspaces`) for backfill/skill re-sync and a `dev:vm` mise task, removes now-dead code (`bootstrap_user_session`, `write_user_root_docs` — INDEX/GUIDE are symlinks into `_system`), and renames the misleading `get_user_connected_integrations` → `get_user_integration_records`. Verified end-to-end against the dockered backend — registration provisioning, session creation, and integration/todos/gaia-tasks sync all materialize correctly on disk — plus a 24-case unit suite; lint and type-check are clean.

### DIFF
--- a/apps/api/app/agents/workspace/system_files.py
+++ b/apps/api/app/agents/workspace/system_files.py
@@ -43,8 +43,8 @@ class SystemFile(NamedTuple):
     body: str
 
 
-# Static docs that anchor the workspace. Paths match what write_user_root_docs /
-# the gaia-tasks + user-todos materializers write, so reads stay consistent.
+# Static docs that anchor the workspace. Paths match what the skill catalog /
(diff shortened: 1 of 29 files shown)


Dhruv-Maradiya:
@coderabbitai pls reivew

PR #768, opened 2026-06-14
refactor(timezone): unify timezone handling behind canonical modules

Dhruv-Maradiya:
Introduces an always-valid `Timezone` value object on the API (`app/utils/timezone.py`) and a branded `Timezone` type on the web (`src/lib/timezone.ts`) as the single sources of truth, making illegal timezone states unrepresentable. All cron, reminder, workflow, scheduler, calendar, and OAuth timezone logic is routed through them, removing the duplicate ad-hoc parsing (`parse_timezone`, `get_user_timezone`, `get_user_time`, `browserTimezone`) that caused silent UTC drift in scheduled tasks. The refactor cleanly separates the **home timezone** (drives the agent's "now", display, and notifications) from the **schedule timezone** (the wall-clock zone a specific cron/reminder/event fires in). Cron math now interprets expressions in the schedule's own zone instead of falling back to UTC. Unit tests across timezone, cron, reminder, scheduler, workflow models, and OAuth dependencies are updated and expanded to cover the new contract.


Dhruv-Maradiya:
@coderabbitai pls review

PR #775, opened 2026-06-16
feat(chat): queue button + MiniMax M3 hardening, model-thinking tool cards & executor reliability

aryanranderiya:
## What

This branch started as the **composer Queue button** and grew to carry the **paid-model (MiniMax M3) rollout hardening** and the chat-side polish that came with it: model-thinking display, executor delivery reliability, comms relay correctness, and tool-card/markdown rendering.

10 commits, ~45 files. Grouped by area below.

## 1. Composer Queue button + queued-message UX (original feature)

When a message is sent while the active conversation's stream is **still open** (after the initial "on it", before the SSE closes), it's held in the pending-stream queue instead of sent immediately. This surfaces it:

- Composer button morphs **Send → Queue** (clock icon + text morph) during the held window; **Stop** is unchanged. Detected reactively via `streamingConversationId === activeConversationId && !isMainResponseStreaming`.
- The held user message renders **greyed-out** (new `"queued"` status) and flips to normal the moment the queue dispatches it. Queued bubbles show a "Queued" label instead of the date/action row.
- Fix: a queued message no longer hijacks the loading indicator.

## 2. Paid model = MiniMax M3 via OpenRouter (hardening)

- Per-plan routing: **free → Gemini, paid → MiniMax M3** (OpenRouter), set on the comms configurable and inherited by the executor + subagents.
- Swapped the OpenRouter client to **`ChatOpenRouter`** so reasoning tokens are exposed (`ChatOpenAI` dropped them); reasoning effort + provider routing are configurable. Pinned to the first-party `minimax` provider lane to dodge reseller 429/400s, and capped output tokens under the shared context window.
- **Array tool-arg fix (root cause + verified):** `retrieve_tools.exact_tool_names` was `list[str] | None`, which renders as an `anyOf: [array, null]` JSON schema. MiniMax M3 cannot populate an array inside that nullable union and always sent `exact_tool_names: []`, breaking tool binding in a loop. Made it a non-nullable `list[str]`. Confirmed against a real M3 call (`[]` to `["GMAIL_CREATE_EMAIL_DRAFT"]`).

## 3. Model thinking in tool cards

- Stream the model's reasoning interleaved with tool steps (subagent + executor), parse a new `reasoning` stream event, render collapsible **Thinking** rows in the unified tool card at root and subagent levels, and persist it.
- Consecutive reasoning entries coalesce into one Thinking block (no chronology break across real tool calls).

## 4. Executor delivery + tool-card reliability

- Reliable executor delivery; follow-up actions no longer leak into the response (marked silent in the metadata stream consumers actually read).
- **Redirect dedupe:** a same-turn "stop X, do Y" (`cancel_executor` + `call_executor`) raced the cancel and queued Y, rendering it as a second tool card. `call_executor` now waits out an in-flight cancel of the lock holder and runs live, so each turn keeps one unified tool card.
- `cancel_executor` gets its own `CancelCircleHalfDotIcon`. Empty-bubble flash on slow start fixed; `formatToolName` made null-safe.

## 5. Comms relay correctness

- Comms must relay every piece of executor output in the right format, and **mirror the executor's own structure/priority** rather than re-triaging or fabricating priority tiers.

## 6. Tool-card / markdown rendering

- Tool output renders **markdown for skill files**: strips the `read` tool's line-number prefixes for `.md` reads so `#`/`##` parse correctly.
- **`CompactMarkdown` now reuses the canonical `MarkdownRenderer`** (Streamdown: GFM, tables, syntax-highlighted code, math, links) with a compact heading/list scale, instead of a second `react-markdown` setup — and gets a proper descending `h1..h6` hierarchy (headings were previously all the same size).
- Follow-up-actions prompt rewritten to be strictly **user-facing** (no tool/function names, phrased in the user's words) with a **conversion** lens; runs on `gemini-3.1-flash-lite`.

## Notes

- Queueing is purely frontend hold-and-dispatch; no backend protocol change.
- Voice-mode investigation (executor answers spoken via `voice_tts` but not always bubbled) was done separately and is **not** in this PR.


aryanranderiya:
### 🧹 Staging Preview · Torn down

| | |
|---|---|
| **Status**  | 🧹 Torn down |
| **Commit**  | [`a721930`](https://github.com/theexperiencecompany/gaia/commit/a7219303b7895cbdf4190ace9f21e7395bfaa030) |
| **Logs**    | [Dozzle](https://gaia-home-server.taila76294.ts.net/logs?filter=name%3Dgaia-pr-775-) |
| **Updated** | 2026-06-20 16:07 UTC |

<sub>Updated automatically by [previewctl](https://github.com/theexperiencecompany/gaia-previewctl) for PR #775.</sub>

PR #787, opened 2026-06-20
fix(download): resolve desktop release via API so buttons download directly

aryanranderiya:
## Problem

Going to `/download` and clicking **Download for Windows** (or Mac/Linux) dumps the user on the raw GitHub releases page instead of downloading the binary — surfaced by a user on a university network.

**Root cause (deterministic, not flaky):** the buttons resolve the latest desktop release client-side from the GitHub API with `?per_page=10`. Every monorepo app publishes into one releases feed, so the 10 most recent releases are all `web`/`api`/`cli`/`bots`/`mobile` and the newest `desktop-*` tag sits at index **[10]** — just past the cutoff. Resolution returns `null` and every button falls back to `releases?q=desktop&expanded=true`.

```
[0] web-v0.20.0 ... [9] mobile-v0.5.0   ← per_page=10 sees only these
[10] desktop-v0.3.0                       ← the desktop release, out of reach
```

A second latent failure: the client hits GitHub **unauthenticated** (60 req/hr **per IP**), so a shared/NAT'd network (office, university) can exhaust it and break downloads for everyone behind it.

## Fix

Move release resolution into our API and drive the UI from the **real** published assets.

### Backend (`apps/api`)
- New public `GET /desktop/releases/latest` — fetches GitHub with `per_page=100`, returns the newest non-draft `desktop-*` tag + its asset list.
- Cached in Redis (30 min) via `@Cacheable`; uses `GITHUB_TOKEN` (5000/hr) so shared IPs can't exhaust the limit.
- Added `/api/v1/desktop/releases` to the WorkOS auth allowlist — scoped so the authed `/desktop/tool-result` bridge stays protected.

### Frontend (`apps/web`)
- `usePlatformDetection` now calls our API and resolves each platform/arch URL by **matching real asset names**. A missing binary resolves to `null` and is simply not offered — never a 404. Falls back to the GitHub releases page if the lookup fails.
- Replaced the Mac-only chip chooser + separate Windows/Linux buttons with **one** arch-aware `DesktopDownloadButton(os)` that shows an x64/ARM picker when >1 binary is published and a single button otherwise. `SettingsMenu`'s hook API is unchanged.

## Result (verified against the live `desktop-v0.3.0` release)

| Platform | Behaviour |
|---|---|
| macOS | Chooser: Intel + Apple Silicon |
| Linux | Chooser: x64 + ARM64 _(previously x64-only)_ |
| Windows | Single x64 button — ARM64 hidden (not published), auto-becomes a chooser once built |

## Out of scope (separate follow-up)
`electron-builder.yml` lists Windows arm64 as a target, but `desktop-v0.3.0` ships no `GAIA-arm64.exe`. The UI auto-shows the Windows chooser the moment that artifact is published — but the desktop **build/CI** needs to actually produce & upload it.

## Verification
- `nx lint api` ✓, mypy on changed files ✓
- `nx run web:type-check` ✓, biome on changed files clean ✓
- Simulated full resolution (backend selection + frontend asset mapping) against the live GitHub release — all targets resolve correctly, Windows ARM correctly hidden.

## Testing notes
No tests added (per repo convention). To verify manually: load `/download` and confirm each platform downloads the binary directly; check `GET /api/v1/desktop/releases/latest` returns `desktop-v0.3.0` + assets without auth.

### DIFF
--- a/apps/api/app/api/v1/endpoints/desktop.py
+++ b/apps/api/app/api/v1/endpoints/desktop.py
@@ -1,8 +1,11 @@
-"""Desktop tool bridge endpoints.
+"""Desktop app endpoints.
 
-The Electron app POSTs results of desktop-executed tool actions here; the
-endpoint validates ownership against the pending-request key and relays the
-payload to the awaiting agent tool over Redis.
+Two concerns share the ``/desktop`` prefix:
+
+- the tool bridge (authed): the Electron app POSTs results of desktop-executed
+  tool actions here, and the endpoint relays them to the awaiting agent tool;
+- release distribution (public): the marketing download page resolves the latest
+  desktop binary so its buttons link straight to the right platform/arch asset.
 """
 
 from typing import Annotated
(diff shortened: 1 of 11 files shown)


aryanranderiya:
### CodeRabbit review — both items verified, no code change needed

**1. `response_model=` on `GET /desktop/releases/latest`** — not applied. Per `apps/api/CLAUDE.md`, when a handler returns a Pydantic model directly (this returns `DesktopReleaseResponse`), the return annotation already defines the schema and adding `response_model=` is flagged redundant by **SonarQube S8409**. The sibling `tool_result` handler in the same file follows the same no-`response_model` pattern. (SonarCloud already passed without it.)

**2. "Missing `ButtonGroup` import" in `DownloadPage.tsx`** — false positive. `ButtonGroup` is imported on line 3 (`import { Button, ButtonGroup } from "@heroui/button";`) and `nx run web:type-check` passes, which it would not if the name were unresolved.

Pushed in `e1ae512`: review feedback (constants moved into `app/constants/`) and the failing **"Types live in dedicated *.types.ts files"** gate (platform types moved to `usePlatformDetection.types.ts`).

PR #810, opened 2026-06-21
feat(analytics): implement subscription analytics sync and user prope…

Vinitthakkar45:
…rties update


apps/api/app/services/analytics_service.py
```diff
@@ -33,6 +34,22 @@ class AnalyticsEvents:
     SUBSCRIPTION_FAILED = "subscription:failed"
 
 
+# Subscription person-property values. Kept in sync with the frontend, which
+# mirrors the same `plan` / `is_subscribed` values onto the person each session
+# (apps/web/src/features/pricing/hooks/useSubscriptionAnalyticsSync.ts).
+# StrEnum members are real strings, so they serialize to PostHog unchanged.
+# Add new tiers (e.g. MAX, ENTERPRISE) here — no new module-level constants.
+class SubscriptionPlan(StrEnum):
```
aryanranderiya: we alreayd have subscription plan enum models in the codebase. reuse that. along with status. so that updating plans doesn't drift

apps/web/src/features/pricing/hooks/useSubscriptionAnalyticsSync.ts
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Keeps PostHog person properties in sync with the user's authoritative
+ * subscription status so dashboards can segment pro vs free users.
+ *
+ * The backend is the source of truth for *becoming* a paying customer (it fires
+ * `subscription:activated` only after a verified Dodo webhook). This hook mirrors
+ * the resulting state onto the person each session, which also self-heals on
+ * downgrade/churn — something a one-shot activation event cannot do.
+ */
+export const useSubscriptionAnalyticsSync = (): void => {
```
aryanranderiya: why is this needed? seems like the hook is just taking status, and remapping it to 2 properties (1 same is subscribed property).

apps/api/app/services/analytics_service.py
```diff
@@ -33,6 +34,22 @@ class AnalyticsEvents:
...
+    PRO = "pro"
+    FREE = "free"
+
+
+class SubscriptionStatus(StrEnum):
```
aryanranderiya: here too we probably have a pre-existing model that can be re-used.

aryanranderiya (changes requested):

apps/api/app/services/analytics_service.py
```diff
@@ -248,23 +265,48 @@ def track_subscription_event(
...
+            "plan": SubscriptionPlan.PRO,
+            "is_subscribed": True,
+            "subscription_status": SubscriptionStatus.ACTIVE,
+        }
+        if event_type == AnalyticsEvents.SUBSCRIPTION_ACTIVATED:
+            properties["subscription_activated_at"] = datetime.now(UTC).isoformat()
+        return properties
+    if event_type == AnalyticsEvents.SUBSCRIPTION_CANCELLED:
+        return {"subscription_status": SubscriptionStatus.CANCELLED}
+    if event_type == AnalyticsEvents.SUBSCRIPTION_EXPIRED:
```
aryanranderiya: shit code with if else if else. clean it up

aryanranderiya (changes requested):

aryanranderiya (approved):

PR #811, opened 2026-06-21
fix(sandbox): make JuiceFS mount reliable + fast (cold 76–104s → ~31s, warm ~2s)

aryanranderiya:
## Problem
The agent's `bash`/`write`/`edit` were excruciatingly slow. Root-caused live in prod (Grafana/Loki + SSH + reproducing real E2B sandbox mounts against the prod CockroachDB + R2). The `ls` itself is milliseconds — **the cost is the per-user JuiceFS *mount* on a cold sandbox acquire.**

Measured decomposition of one mount, Oregon sandbox → Frankfurt metadata:
- connect (TLS verify-full handshake) ~1.2s · read volume format from R2 ~0.7s · **SQL `xorm` schema-sync ~9.5s** · session+FUSE ~1s ≈ **~13s/mount**
- `mount.sh` does **three** mounts (user+skills+system) → sequentially **~43s**
- on the old code, OOM (600MB buffers in a 1 GB sandbox) + the 10s `--background` FATAL + canary-stale recreate loop pushed real turns to **76–104s**, with 3–9 re-acquires/turn
- **R2 is ~40ms from the sandbox — ruled out.** The cost is `round-trips × distance`.

## What this PR changes (all validated in real prod)
1. **OOM fix** — `--buffer-size` 600→128MB primary, 300→32MB skills/system (+cache trims). The 1 GB sandbox OOM-killed the mount ~2/3 of the time. Verified: 600MB fails 2/3; 128MB reliable (~170–260MB used).
2. **Foreground mount** — drop `--background` (self-aborts at a hardcoded 10s); launch detached-foreground (`setsid … &`), `wait_mounted` is the real gate.
3. **Parallel mounts** — launch all 3 daemons up front, then wait+bind each (was launch+wait ×3). Same-conditions A/B (interleaved, seeded user): **sequential 43.0s → parallel 31.3s median** (~28% faster, far less variance).
4. **Idle-pause 60→300s** — keep the sandbox warm across a session so only the *first* acquire mounts; later turns are cache hits.

## Measured numbers — current prod vs after this PR
| Scenario | Current prod | After this PR |
|---|---|---|
| **Cold hit** (first action, or after >5min idle → paused sandbox remounts) | **76–104s** | **~31s** (mount.sh parallel median 31.3s; full acquire ~33s) |
| **Warm hit** (within the active window) | ~1.6–3s | **~1.6–3s** (unchanged — it's a cache hit — but hit far more often: idle-pause 60→300s) |

Reference measurements: old cold full acquire 70.2s / 71.2s · warm acquire 1.6s, warm acquire+ls 3.0s · new mount.sh parallel 31.3s vs old sequential 43.0s.

## Why ~31s is the code floor — and what fixes cold for real
The residual ~31s is **3× contended SQL schema-sync × the sandbox↔metadata network distance** — irreducible in code. Proven: **at 0ms RTT the *same* mount is ~1.2s** (vs ~13s/mount at 150ms; ×3 contended ≈ 31s). So distance, not the engine, is the wall.

**The real cold-case fix is co-location: run E2B sandboxes in EU, next to Cockroach + R2 + the host → mount ~1.2–3s, with zero code/sync/durability tradeoff. The user has contacted E2B support to enable an EU region.** Once that lands, even a true cold hit is ~3s.

## Notes
- **No Redis.** All numbers are with the existing **CockroachDB** metadata engine; this PR does not change the engine. (Redis would cut the schema-sync to ~5.8s@150ms but trades away durability — metadata loss = filesystem loss — so it's intentionally not used.)
- `_run_mount_script` ships the API's current `mount_juicefs.sh` at runtime, so the mount changes take effect on the next sandbox **without an E2B template rebuild**.
- A prewarm endpoint was prototyped and **dropped**: without co-location it only marginally hides the cold mount, server-side it's ~what `acquire_sandbox` already does at the coding tool call, and warming on conversation-open would waste sandboxes on non-coding chats (quota risk). Keep-warm (idle-pause) + co-location cover it cleanly.
- **Follow-up (not in this PR):** E2B-EU co-location (in progress with E2B support); optional dedicated (non-serverless) Cockroach to remove load-driven variance.

### DIFF
--- a/apps/api/app/config/settings.py
+++ b/apps/api/app/config/settings.py
@@ -280,7 +280,15 @@ class ProductionSettings(CommonSettings):
     # ----------------------------------------------
     E2B_API_KEY: str
     E2B_TEMPLATE_ID: str  # gaia-coder template ID (run scripts/build_e2b_template.py)
-    E2B_SANDBOX_IDLE_PAUSE_SECONDS: int = 60
+    # Idle window before a sandbox is paused. A paused sandbox must resume +
+    # re-mount JuiceFS on the next turn, and the cold JuiceFS mount is the single
+    # most expensive step in an acquire (the metadata engine is remote). At 60s,
+    # any think-gap between turns paused the sandbox and made the *next* `bash`
+    # pay a full remount. 300s keeps the sandbox warm across normal conversation
+    # gaps so back-to-back turns reuse a live mount. Trade-off: more concurrently
+    # live sandboxes vs the E2B quota — the scalable fix is the warm pool
+    # (E2B_WARM_POOL_TARGET_RATIO), still a follow-up.
+    E2B_SANDBOX_IDLE_PAUSE_SECONDS: int = 300
     E2B_DEFAULT_BASH_TIMEOUT: int = 120
     E2B_SANDBOX_EVICT_DAYS: int = 14
     E2B_WARM_POOL_TARGET_RATIO: float = 2.0  # Phase 2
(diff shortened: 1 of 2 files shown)


aryanranderiya:
Added a **prewarm endpoint** (`POST /sessions/{conv_id}/prewarm`): fire-and-forget background acquire+mount, 202, Redis-NX-deduped (≤1 warm in flight per user → no quota pile-up). Validated in prod: dedupe works; after a prewarm the next `acquire+ls` drops **71s → 3.0s** (warm hit). The cold mount now overlaps conversation-open + agent reasoning instead of blocking the first tool call.

**Remaining (not in this PR):**
- *Frontend hook* — call `POST /sessions/{conv}/prewarm` on conversation focus / when coding is likely (it's opt-in by design so we don't warm a sandbox on every non-coding chat).
- *Co-location (infra, the real cold-case fix)* — run E2B sandboxes in EU next to Cockroach+R2. Measured: at 0ms RTT the same mount is ~1.2s vs ~31s cross-region; the residual ~31s is `3× contended SQL schema-sync × distance`, irreducible in code. With co-location the first bash becomes ~instant even without prewarm.
- *Optional*: dedicated (non-serverless) Cockroach to remove the load-driven variance (real turns hit 76–104s under load vs ~31–43s quiet); keeps durability.

PR #827, opened 2026-06-25
feat(workflow): deliver results into linked platform chats as real messages

aryanranderiya:
## What & why

When a workflow finishes, its result reached the user on messaging platforms **only** through the proactive notification pipeline — wrapped in templated "workflow done" header/footer chrome, never written into the platform's own conversation thread, and worded differently from the saved copy. So on Telegram the message didn't sound like GAIA and couldn't be continued, and it wasn't really "in the chat".

This delivers a workflow result the same way a normal reply is delivered: as a **real, persisted, multi-bubble message in GAIA's voice**, into each of the user's linked + preferred platform conversations — while keeping the existing `workflow_system` copy for the web.

## Changes

- **`result_delivery.py`** — new `_deliver_workflow_to_platforms` (+ `_preferred_bot_platforms`, `_post_workflow_message`): fans out to platforms the user has **linked AND left enabled** in their notification channel preferences (reuses `fetch_channel_preferences`), persists the result into each platform's session conversation, and publishes it split into bubbles on the message-break sentinel.
- **`comms_prompts.py`** — `PLATFORM_DELIVERY_NOTE` now tells comms it's a **workflow** result the user hasn't seen, to surface **every** result in full, and to split into multiple bubbles in GAIA's voice.
- **`notifications.py`** — completion notification scoped to **in-app only** (no external chrome push, no `View Results` link, no result payload); failure copy humanized.
- **`constants/notifications.py`** — humanized `WORKFLOW_DONE_COPY`, no robotic timestamp.
- **`external.py`** — removed the now-dead `workflow_execution` branch + `View full results` footer.
- **`formatters.ts`** — **Telegram only**: render markdown blockquotes as `<blockquote>` / `<blockquote expandable>` (latest Telegram formatting) instead of stripping them. No other platform touched.
- **Tests** — updated `test_notification_channels.py` + `test_workflow_tasks.py` to the new contract.

## Verification

- ruff + mypy clean (686 files); Biome + tsc clean; 75 formatter tests + updated unit tests pass.
- Real workflow run through the ARQ worker → 5-bubble, GAIA-voiced, chrome-free envelope delivered to Telegram/Discord/WhatsApp, persisted into each platform conversation + the `workflow_system` thread, with the completion notification delivered in-app only (no double-send). Telegram formatting (incl. blockquotes) confirmed rendering in the live client.

### DIFF
--- a/apps/api/app/agents/core/background/result_delivery.py
+++ b/apps/api/app/agents/core/background/result_delivery.py
@@ -23,24 +23,21 @@ from langsmith import traceable
 from app.agents.core.background.comms_narrator import narrate_executor_result
 from app.agents.core.background.executor_capture import drain_executor_tool_data
 from app.agents.core.background.session import ExecutorRun
+from app.agents.core.background.workflow_platform_delivery import (
+    deliver_workflow_result_to_platforms,
+)
 from app.agents.core.nodes.follow_up_actions_node import generate_follow_up_actions
-from app.constants.general import NEW_MESSAGE_BREAKER
 from app.constants.log_tags import LogTag
 from app.core.websocket_manager import websocket_manager
 from app.db.mongodb.collections import conversations_collection
 from app.models.chat_models import (
-    BOT_CONVERSATION_SOURCES,
     ConversationSource,
     MessageModel,
     UpdateMessagesRequest,
 )
 from app.models.message_models import ReplyToMessageData
-from app.services.bot_service import BotService
 from app.services.conversation_service import update_messages
-from app.services.outbound_delivery import publish_outbound_message
-from app.services.platform_link_service import PlatformLinkService
 from app.services.platform_message_service import deliver_message_to_platform, is_bot_platform
-from app.utils.notification.channel_preferences import fetch_channel_preferences
 from shared.py.wide_events import log
 
 
@@ -82,7 +79,7 @@ async def deliver_result(
         return await _narrate_and_deliver(
             run, result_text, result_type, attach_tool_data, returned_note
         )
-    except Exception as e:  # noqa: BLE001 — delivery is best-effort, never propagates
+    except Exception as e:  # delivery is best-effort, never propagates
         log.error(f"{LogTag.AGENT} Background notification delivery failed", error=str(e))
         return None, None
 
@@ -124,7 +121,7 @@ async def persist_cancelled_run(run: ExecutorRun) -> None:
             ),
             user=run.user,
         )
-    except Exception as e:  # noqa: BLE001 — best-effort save of a stopped run
+    except Exception as e:  # best-effort save of a stopped run
         log.error(f"{LogTag.AGENT} Failed to save cancelled executor cards", error=str(e))
         return
 
@@ -233,7 +230,7 @@ async def _narrate_and_deliver(
         # messaging-platform conversations as normal bot messages (GAIA's voice,
         # no notification chrome). The in-app badge below is a web-only heads-up.
         if result_type != "error" and run.workflow_notify_on_completion:
-            await _deliver_workflow_to_platforms(
+            await deliver_workflow_result_to_platforms(
                 user=run.user,
                 user_id=user_id,
                 notification_text=notification_text,
@@ -242,6 +239,7 @@ async def _narrate_and_deliver(
             msg_type=result_type,
             workflow_id=run.workflow_id,
             workflow_title=run.workflow_title,
+            conversation_id=run.conversation_id,
             user_id=user_id,
             message_id=bot_message.message_id,
             notify_on_completion=run.workflow_notify_on_completion,
@@ -321,7 +319,7 @@ async def _safe_inline_follow_ups(
             user_msg_content=user_msg_content,
             user_id=user_id,
         )
-    except Exception as e:  # noqa: BLE001 — follow-ups are best-effort
+    except Exception as e:  # follow-ups are best-effort
         log.error(
             f"{LogTag.AGENT} deliver_result: failed to generate follow-up actions",
             error=str(e),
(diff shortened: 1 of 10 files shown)


apps/api/app/agents/core/background/result_delivery.py
```diff
@@ -418,14 +435,117 @@ async def _generate_and_push_follow_ups(
...
+    """Resolve which messaging platforms a workflow result should reach: those the
+    user has linked AND left enabled in their notification channel preferences."""
+    try:
+        linked = await PlatformLinkService.get_linked_platforms(user_id)
+        prefs = await fetch_channel_preferences(user_id)
+    except Exception as e:  # noqa: BLE001 — proactive side channel, never fatal
+        log.error(f"{LogTag.AGENT} workflow platform delivery: target lookup failed", error=str(e))
+        return []
+
+    targets: list[tuple[ConversationSource, str]] = []
```
aryanranderiya: maybe this code can be cleaner it doesn't look good, maybe can be simplified? more pythonic? cleaned up?

apps/api/app/agents/core/background/result_delivery.py
```diff
@@ -418,14 +435,117 @@ async def _generate_and_push_follow_ups(
...
+            continue
+        if not prefs.get(platform_value, True):
+            continue  # channel disabled in the user's notification preferences
+        platform_user_id = info.get("platformUserId")
+        if platform_user_id:
+            targets.append((source, str(platform_user_id)))
+    return targets
+
+
+async def _post_workflow_message(
```
aryanranderiya: is this the most correct place to have this helper function? in terms of tech debt

apps/api/app/constants/notifications.py
```diff
@@ -45,25 +45,24 @@
...
+# {title} is the workflow name. One pair is picked per run so repeats don't read
+# like a robot. The workflow's full result is delivered as real messages straight
+# into the user's chat, so bodies can warmly point there ("in your chat") and must
+# never reference a link or button (there is none).
 WORKFLOW_DONE_COPY: tuple[tuple[str, str], ...] = (
-    ("just wrapped up {title} 🙌", "all done at {time}"),
-    ("ok, {title} is done!", "finished at {time}, all yours"),
-    ("just finished {title} 🎉", "wrapped it up at {time}"),
-    ("sorted {title} for you", "all good as of {time}"),
+    ("sorted {title} for you", "dropped it all in your chat 🙌"),
```
aryanranderiya: add a couple more options these might not be enough. and i don't like the dropped it in chat and xyz is in your chat. think from how it'll look like from the user's perspective

apps/api/app/services/workflow/notifications.py
```diff
@@ -1,90 +1,55 @@
...
 async def send_workflow_completion_notification(
     *,
     workflow_id: str,
     workflow_title: str,
-    conversation_id: str,
     user_id: str,
-    result_text: str,
 ) -> None:
-    """Send the "workflow done" notification, splitting the result into bubbles.
+    """Send the in-app "workflow done" heads-up (web only).
```
aryanranderiya, line 38: why is it setup like this? like why is this logic decoupled to how we send the workflow completion notification on all the different bots? like i get how it's just for a nudge, but these 2 different paths are confusing and could contribute to tech debt.

apps/api/app/services/workflow/notifications.py
```diff
@@ -1,90 +1,55 @@
...
     try:
         await notification_service.create_notification(
             NotificationRequest(
                 user_id=user_id,
                 source=NotificationSourceEnum.WORKFLOW_COMPLETED,
                 type=NotificationType.SUCCESS,
-                content=NotificationContent(
-                    title=title,
-                    body=body,
-                    actions=[
```
aryanranderiya, line 66: why did we remove the actions? shouldn't this be shown in the web app?

apps/api/app/utils/notification/channels/external.py
```diff
@@ -57,26 +61,11 @@ async def transform(self, notification: NotificationRequest) -> dict[str, Any]:
         before sending — no platform-specific markdown is produced here.
         """
         content = notification.content
-        rich = content.rich_content or {}
         app_url = settings.FRONTEND_URL.rstrip("/")
         title = content.title or ""
         body = content.body or ""
-        header = _join_nonempty(f"**{title}**" if title else "", body)
```
aryanranderiya, line 64: why should we remove this? please explain

apps/api/app/agents/core/background/result_delivery.py
```diff
@@ -418,14 +435,117 @@ async def _generate_and_push_follow_ups(
...
+        )
+
+
+async def _preferred_bot_platforms(user_id: str) -> list[tuple[ConversationSource, str]]:
+    """Resolve which messaging platforms a workflow result should reach: those the
+    user has linked AND left enabled in their notification channel preferences."""
+    try:
+        linked = await PlatformLinkService.get_linked_platforms(user_id)
+        prefs = await fetch_channel_preferences(user_id)
+    except Exception as e:  # noqa: BLE001 — proactive side channel, never fatal
```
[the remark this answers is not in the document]
aryanranderiya: Done — though the root cause was a bit different from "narrow the except". `BLE001` is in ruff's **global** ignore list (`pyproject.toml` → `[tool.ruff.lint] ignore`, "Allow `try/except Exception` — many legitimate top-level safety nets"), so these `# noqa: BLE001` were **redundant**, not load-bearing. Verified empirically: stripping every `# noqa: BLE001` from the file keeps `ruff check` green.

So I removed the suppression syntax entirely (Sonar-clean, and consistent with the no-blanket-noqa guideline) while keeping the *rationale* as a plain trailing comment, e.g. `except Exception as e:  # best-effort save of a stopped run`. The two lines you flagged (481/535) moved into the new `workflow_platform_delivery.py` and lost the `noqa` there too.

apps/api/app/services/workflow/notifications.py
```diff
@@ -108,10 +73,10 @@ async def send_workflow_failure_notification(
                 source=NotificationSourceEnum.WORKFLOW_FAILED,
                 type=NotificationType.ERROR,
                 content=NotificationContent(
-                    title=f"Workflow Failed: {workflow_title}",
+                    title=f"couldn't finish {workflow_title}",
                     body=(
-                        f"Your workflow '{workflow_title}' encountered an error "
-                        f"and could not complete."
+                        f"'{workflow_title}' ran into an error and stopped partway. "
+                        "nothing was lost, you can re-run it whenever you're ready."
```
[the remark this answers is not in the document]
aryanranderiya: Done — dropped the over-promise. The failure body now reads: `'{workflow_title}' ran into an error and stopped partway. you can re-run it whenever you're ready.` (no more "nothing was lost").

apps/api/tests/unit/workers/test_workflow_tasks.py
```diff
@@ -970,36 +971,32 @@ async def test_user_message_has_selected_workflow(self):
...
-            patch(
-                "app.services.workflow.notifications.notification_service",
-            ) as mock_notif,
-            patch(
-                "app.services.workflow.notifications.get_user_by_id",
-                new_callable=AsyncMock,
-                return_value={"timezone": "Asia/Kolkata"},
-            ),
-        ):
+    async def test_completion_notification_is_inapp_only(self):
```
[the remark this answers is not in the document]
aryanranderiya: Done — added `-> None`.

libs/shared/ts/src/bots/utils/formatters.ts
```diff
@@ -220,6 +220,45 @@ function stashTables(text: string, hold: (html: string) => string): string {
...
+  let i = 0;
+  while (i < lines.length) {
+    if (!isQuote(lines[i] ?? "")) {
+      out.push(lines[i] ?? "");
+      i += 1;
+      continue;
+    }
+    const quoted: string[] = [];
+    while (i < lines.length && isQuote(lines[i] ?? "")) {
+      quoted.push((lines[i] ?? "").replace(/^&gt;[ \t]?/, ""));
```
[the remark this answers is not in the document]
aryanranderiya: Done — `isQuote` and the marker-strip now share a single `/^[ \t]{0,3}&gt;[ \t]?/` regex, so blockquotes with up to 3 leading spaces (CommonMark's limit) are wrapped instead of rendering literal `&gt;`. Added a test: `wraps an indented blockquote (up to 3 leading spaces)`.

apps/api/app/agents/core/background/result_delivery.py
aryanranderiya: Cleaned up — and moved it into the new `workflow_platform_delivery.py` while I was at it. The three `continue` guards collapsed into one condition:

```python
for platform_value, info in linked.items():
    source = ConversationSource.coerce(platform_value)
    platform_user_id = info.get("platformUserId")
    if (
        source is not None
        and source in BOT_CONVERSATION_SOURCES
        and prefs.get(platform_value, True)
        and platform_user_id
    ):
        targets.append((source, str(platform_user_id)))
```

(`source is not None` stays so mypy narrows it for the tuple; `in BOT_CONVERSATION_SOURCES` already filters unknown/non-bot platforms.)

apps/api/app/agents/core/background/result_delivery.py
aryanranderiya: Agreed, this was the wrong home. `result_delivery.py` was ~690 lines mixing two concerns: generic executor-result delivery (narrate → persist → route over one transport) and workflow-specific platform fan-out. I extracted the whole fan-out unit — `deliver_workflow_result_to_platforms` + `_preferred_bot_platforms` + `_post_workflow_message` — into a dedicated **`app/agents/core/background/workflow_platform_delivery.py`** (with a module docstring explaining how it relates to the in-app badge). `result_delivery.py` now just imports the one entry point and drops back to ~570 lines. No import-topology change (it already imported those same services), so no new cycles.

apps/api/app/constants/notifications.py
aryanranderiya: Reworked the copy. Two changes:

1. **Dropped the "in your chat" framing.** It assumes the user has an external messaging chat, but a web-only user doesn't — for them the result lives in a GAIA conversation they reach via the (now restored) "View Results" button. So the bodies are channel-agnostic now.
2. **More variety** — 6 options instead of 4.

```
("sorted {title} for you",     "it's all ready whenever you are 🙌"),
("{title} is done",            "had a proper look — everything's ready for you"),
("just wrapped up {title}",    "pulled it all together, take a peek"),
("handled {title} for you",    "all done end to end, give it a look"),
("finished {title}",           "got everything ready for you to check out"),
("{title}: all set",           "took care of it, here's what I found"),
```

Still first-person and casual, just no longer claiming a specific place.

apps/api/app/services/workflow/notifications.py
aryanranderiya: They look like two paths because they hit **two different surfaces**, and that separation is actually the fix this PR makes — not new debt:

- **`deliver_workflow_result_to_platforms`** (in `workflow_platform_delivery.py`) posts the *actual result* into the user's real Telegram/WhatsApp/Discord/Slack chat as a natural GAIA message — persisted to a conversation + sent via the outbound bot queue. It *is* the result, continuable like any other message.
- **`send_workflow_completion_notification`** (this file) is just a **web in-app badge** via the notification framework — a heads-up, not the result.

Before this PR these were *one* path: the notification framework fanned out to inapp **+** external channels, and the external adapter reconstructed the result bubbles + a "View full results" link so it could shove the result to bots. That meant a real chat message was being delivered *through notification chrome*. That's the coupling we're removing — bot users should get a real message, not a notification.

So collapsing the two back together would re-introduce exactly that coupling. They're decoupled on purpose because the badge is intentionally **not** the result — it's a nudge, on a surface (web notification center) the chat delivery doesn't reach. I added a module docstring to `workflow_platform_delivery.py` spelling this out so the next reader doesn't trip on it.

apps/api/app/services/workflow/notifications.py
aryanranderiya: Good catch — **restored** the link. The in-app badge now carries a single "View Results" REDIRECT to `/c/{conversation_id}`, so a web user (who has no external chat) still reaches the run's conversation in one tap. I threaded `conversation_id` back through `_dispatch_workflow_notification` → `send_workflow_completion_notification`, and updated the test to assert exactly one redirect action.

I kept the **`rich_content` / result-bubbles removed**, though — that's a different thing from the link:
- I checked the web notification UI and nothing reads `rich_content` (no component references it; `EnhancedNotificationCard` / `NotificationItem` render title + body + `actions` only). So the embedded result bubbles were dead payload on the client.
- The result itself now lives in the conversation the button opens, so duplicating it inside the notification was redundant.

Net: the useful affordance (one-click to results) is back; the dead duplicated payload stays gone.

apps/api/app/utils/notification/channels/external.py
aryanranderiya: That block was the `rich_content["type"] == "workflow_execution"` branch — it reconstructed the workflow's result bubbles plus a "View full results" footer link so the **external adapter** could push the result out to bots. It's removed because workflow results no longer travel through the notification adapter at all: they're delivered as real chat messages by `deliver_workflow_result_to_platforms` (the `workflow_platform_delivery.py` path).

If we kept this branch, the same result would reach the same platforms **twice** — once as a real message, once as a notification-with-link — which is the duplication this PR is undoing. So `transform` now just renders `title`/`body`/`actions` generically for every notification type. The two tests that covered the old branch were updated to `test_rich_content_is_ignored` (asserts the external transform ignores `rich_content` and emits only the title/body part).

PR #841, opened 2026-06-28
feat(agents): model fallback + unify LLM call sites + remove goals feature

aryanranderiya:
## Summary

This started as "fall back to the default model when the selected model fails" and grew into a broader **LLM-client hardening + call-site unification** pass, plus a full **goals-feature removal** and the cleanup that fell out of it.

`116 files changed, +931 / −8301` (net removal — mostly the goals feature).

Three related workstreams:

1. **Model retry + fallback** — the originating feature: retry the selected model on transient errors, then fall back to the default Gemini model on a hard failure (e.g. OpenRouter 402).
2. **One LLM call primitive + unified call sites** — every auxiliary LLM call now goes through the same helpers (`ainvoke_llm` / `ainvoke_structured` / `response.text`), with the retry/fallback baked in.
3. **Remove the goals feature** — obsolete and unused; deleted full-stack without touching todos.

---

## 1. Model retry + fallback

- **Retry-then-fallback policy.** A single call primitive (`ainvoke_llm` / `invoke_llm` in `client.py`) wraps the bound model with `with_retry` (applied *after* `bind_tools`, so the `RunnableRetry` wrapper never has to expose `bind_tools`) and falls back to the default Gemini model on a provider failure. This replaced the node-level `RetryPolicy` (comms / executor / subagents) and **deleted `retry_policies.py`** — one retry mechanism instead of two.
- **Curated exception sets** (`agents/llm/exceptions.py`) against the *real* `openrouter` SDK hierarchy (it does **not** use the `openai` SDK):
  - Fallback set uses the base `OpenRouterError` (covers 402 + any future error type) + `NoResponseError` + `GoogleAPICallError` + stdlib. It is a curated provider-error set, **not** a bare `except Exception` — programming bugs fail loud instead of silently downgrading the model.
  - Retry set enumerates only the *transient* OpenRouter subclasses, so a permanent 402 isn't retried 3× pointlessly.
- **Fail loud when the comms graph can't be built** (`agent.py`) — a `None` graph now raises with a pointer to the startup error instead of surfacing later as `'NoneType' object has no attribute 'astream'`.
- **The default model carries its context-window profile** (`get_default_llm`). LangChain's curated profile registry lags new model releases and has no entry for the current default model; the summarization/compaction middleware express their triggers as a *fraction* of the context window and raise at construction without it — which failed the whole agent-graph build (the root cause of the production `astream` crash). The window is now attached on the default model so any window-needing consumer just works.
- **Regression tests** build the **real production middleware stack and comms graph** (not the degraded test default), so a middleware-construction regression fails in the test instead of shipping green.

## 2. One LLM call primitive + unified call sites

After an exhaustive audit of every LLM call site in `apps/api/app`, all auxiliary calls were routed through the canonical helpers:

- **`ainvoke_structured(schema, prompt, *, label, temperature=0.1, config=None)`** — the single canonical one-shot structured call (binds the schema, invokes via `ainvoke_llm`, returns the validated model). Accepts any LangChain input (string or message list) + an optional run config.
- **Migrated ~12 structured sites onto it** — 10 `with_structured_output` sites (integration inference, onboarding clarify / inbox-triage / writing-style / intelligence, profile card) and 2 `PydanticOutputParser` sites (`profile_extractor`, `follow_up_actions`) switched from prompt-based parsing to the model's native structured output, dropping the `{format_instructions}` boilerplate.
- **Standardized content flattening on `response.text`** — the branch had been flattening `AIMessage` content three different ways (`.text`, a manual content-block join, and `str(response)`). One canonical accessor now.
- **Naming/constants cleanup** — dropped the misleading leading underscore on the cross-module-imported exception sets and moved `LLM_RETRY_MAX_ATTEMPTS` into `constants/llm.py`.

Deliberately left as-is (with reasons): `social_profile_service` and `profanity` keep their `try/except RuntimeError` no-provider guard around `get_default_llm`; `memory/extraction._invoke_structured` keeps its silent-config + graceful-`None` specialization; the main agent graph keeps its intentional raw/streaming invocation.

### Bug fixes (found during the audit)
- **`file_utils`** was `str(response)`-ing the whole `AIMessage` (metadata, `tool_calls`) into the page/image summaries it embeds into ChromaDB → fixed to `response.text`. Its batch summariation (`process_doc.abatch`) now routes through `with_llm_retry` so it gets the same transient-error retry as every other call.

## 3. Remove the goals feature (full-stack)

The goals feature was obsolete and unused. Removed end-to-end while keeping **todos fully intact**:

- **Backend** — deleted the goals endpoint, service, agent tools, models, utils, docstrings; the goals **delegated subagent** (registry category, `oauth_config` entry, `GOALS_AGENT`/`MEMORY` prompts), the `goal_tracking` rate limit, `goals_collection` + its indexes, and the `goal_data` stream type.
- **Todos decoupling** — the only coupling was the now-dead bidirectional goal-node↔subtask sync (`sync_goal_node_completion`, `sync_subtask_to_goal_completion`, `create_goal_project_and_todo`); removed it and the emptied `sync_service`.
- **Frontend (web / mobile / shared)** — deleted goals routes, the feature module, chat goal cards/renderers, headers; removed goal types/fixtures from `@gaia/shared`, the `goal_data` tool renderer, and the goals nav / command-menu / shortcuts / analytics entries.

> Noted residual (out of scope, needs a product call): the always-empty `roadmap` field on the goal model and `get_goal_statistics`'s node-counting read empty data harmlessly; reworking them is a data-model + frontend change.

## 4. Cleanup
- Removed restate-the-obvious comments and fixed stale docstrings (e.g. `file_utils` docstrings documenting a `file_url` param that doesn't exist) — comments/docstrings only, every "why" comment kept.
- Removed a dead-code chain the strict gate surfaced: `email_comprehensive_parser` → `EmailComprehensiveAnalysis` model → `EmailImportanceLevelEnum`.

---

## Testing & verification

- **Real providers** — verified against a genuinely-out-of-credits OpenRouter (real `PaymentRequiredResponseError` → classified `in_fallback_set=True`, `in_retryable_set=False` → fell back, `gemini-3.1-flash-lite` answered, accepted a bound tool schema) and Infisical-injected live Gemini.
- **Real-Gemini smoke test** of the parser→native-structured migration: `FollowUpActions` (string and message-list forms), `UsernameExtraction`, and `_ClarifyQuestionList` all produced valid structured output.
- `nx lint api` + `nx type-check api` (mypy, 679 source files) green; **6611 unit tests pass**; web type-check + lint green.
- **CI is green** — `test-python`, `Dead code (knip + vulture, strict)`, and the quality gate all pass.

## Gotchas / known tradeoffs

- **Silent downgrade** — when OpenRouter is down the whole turn runs on Gemini with no user-facing signal (log only), and a reasoning model → non-reasoning Gemini changes output quality silently. Product decision if we want to surface "using fallback".
- **`openrouter` SDK coupling** — imports internal error classes; the fallback uses the stable base class, only the transient-retry enumeration is version-sensitive.
- **parser → native structured output** for `profile_extractor` / `follow_up_actions` is a mechanism change in a (background and live) path; validated via the real-Gemini smoke test above and consistent with the 10 other sites already on `with_structured_output`.

---

## Post-review hardening (8 follow-up commits)

A deep review + live E2E pass (real API, real providers) over this branch surfaced one critical gap and a set of smaller ones; all are fixed on this branch. Now `128 files changed, +1,234 / −8,701`.

### The critical one: Gemini retry/fallback was inert
The curated exception sets were built from the **`google-api-core`** hierarchy (`GoogleAPICallError`, `ResourceExhausted`, ...), but `langchain-google-genai` 4.x sits on the **google-genai SDK**: 4xx (incl. 429 quota) surfaces as `ChatGoogleGenerativeAIError` and 5xx as `google.genai.errors.ServerError` — neither subclasses the old types. So for the default provider, `with_llm_retry` never retried, `ainvoke_llm` never fell back, and every narrowed catch keyed on the sets (chatbot degrade, memory extraction's never-raises contract, workflow generation) let real provider errors escape. Proven live: a failing Gemini primary raised straight through the fallback path.

**Fix:** sets rebuilt from the actually-raised classes (`ServerError` retryable; `ChatGoogleGenerativeAIError` + `GeminiAPIError` in the fallback set), plus regression tests that assert coverage of the real classes — the old tests simulated failures with the same wrong exception, which is exactly how this shipped green. Verified live: failing Gemini primary → falls back → answers. Note: Gemini 429s are wrapped in `ChatGoogleGenerativeAIError` (status class hidden), so they route to fallback rather than burning retries on permanent 4xx.

### Fallback + client hardening
- Fallback invocations are **retry-wrapped** (`with_llm_retry(fallback)`) — the last-resort call no longer gets a single unlucky shot.
- `ainvoke_llm`/`invoke_llm` accept a **lazy fallback factory**; `create_agent` no longer re-binds the full tool list to the fallback on every turn (binding happens only if the primary fails).
- **Subagents get the same fallback policy** as the main model node (previously an OpenRouter outage silently downgraded the parent but errored `spawn_subagent` steps); `subagent_end` is now always emitted so a failed spawn can't leave the UI row spinning; the max-turns path guards `.text` against reasoning-only content.
- `get_default_llm()` is **cached per temperature** (was rebuilding a `ChatGoogleGenerativeAI` + HTTP client on every auxiliary call — follow-ups, memory extraction, etc.).
- Typed **`LLMNotConfiguredError`** replaces the bare-`RuntimeError` entry in the chatbot set — programming bugs fail loud again; memory extraction also catches **`OutputParserException`** (what the json_schema parser actually raises on malformed output) and now delegates to `ainvoke_structured`.
- `profanity` opts out of retry backoff (`max_attempts=1`) inside its hard 6s budget; `DEFAULT_LLM_TEMPERATURE` constant replaces the four inlined `0.1`s; `init_fallback_llm` and the duplicated fallback-log blocks are gone.

### Consolidation completed (the last stragglers)
- `generate_workflow_prompt` — was still on `PydanticOutputParser` + `{format_instructions}` with an unhandled `parse()`; now `ainvoke_structured`.
- `generate_steps_with_llm` — dropped the always-True `hasattr(llm, "with_structured_output")` guard and the unreachable raw-JSON text-parse fallback; keeps its schema-invalid regeneration loop on `ainvoke_structured`.
- `/mail/ai/compose` — prose JSON contract + `json.loads` over the chatbot path replaced with `ainvoke_structured(ComposedEmailOutput)`; **bonus bug fix:** the endpoint computed user notes but the template had no `{notes}` placeholder — notes were silently discarded, now injected.
- Deleted dead `EMAIL_COMPREHENSIVE_ANALYSIS` prompt and the orphaned `get_format_instructions` + parser pair in `subagent_output.py`.

### Root-cause + consistency fixes
- **`GraphManager.get_graph` no longer swallows init failures into `None`** — raises typed `GraphUnavailableError` naming the graph and the startup log; `agent.py` drops its symptom-level raise; narrator and subagent-runner keep their degrade contracts but log the real cause at error level.
- **One context-window source of truth** — the summarization archive gate read a `_max_tokens` attribute LangChain never sets (always 128K, inconsistent with the 600K trigger); now resolved via the model profile. `WorkspaceCompactionMiddleware` gets `DEFAULT_MAX_TOKENS` explicitly instead of a hardcoded 128K. The remaining model-blind limitation (triggers denominated in the default model's window) is documented at the constant.
- Goals leftovers removed: 7 `/goal` dummy slash commands + `goal_tracking` category on the landing demo, and the orphaned `goal_tracking` icon entry in `@gaia/shared`.

### Verification (this round)
- Live E2E through the real API (`/chat-stream` SSE, agent JWT): chat streaming, memory add/recall, executor delegation, follow-up actions — all green post-change.
- Real-provider proofs: failing-Gemini primary **falls back** (previously raised); failing-OpenRouter primary + lazy tools-bound factory falls back; no-fallback path raises loud.
- Real-Gemini smoke of the migrated paths: workflow prompt generation (valid schedule trigger + cron) and mail compose (notes actually reflected in the body).
- 6,612 unit tests + integration green; `nx lint/type-check api`, web/desktop type-check + lint green; **CI fully green** on the final push.


apps/api/app/agents/llm/client.py
```diff
@@ -281,48 +308,78 @@ def _create_configurable_llm(primary: LLMProvider, alternatives: list[LLMProvide
...
+    messages: LanguageModelInput,
+    *,
+    fallback: Runnable | None = None,
     config: RunnableConfig | None = None,
-) -> BaseMessage:
-    """Invoke LLMs in sequence until one succeeds, returning its response.
+    label: str = "model",
+) -> Any:
+    """The single way to invoke an LLM. Retries transient provider/infra errors
+    (``with_llm_retry``), then falls back to ``fallback`` (if given) on a provider
```
aryanranderiya: you are overly explaining shit inthe docstring

PR #847, opened 2026-06-30
fix(workflow): cap per-workflow conversation (16MB) + raise Pro trigger limits

aryanranderiya:
## Problem

A workflow's executions are persisted into a **single conversation per `workflow_id`** (`get_or_create_workflow_conversation`). Every run appends a trigger message + result via `update_messages`, which does an unbounded `$push`. Over many runs the document grows until it crosses MongoDB's **16MB cap**, after which **every** subsequent run fails at message persistence with `WriteError 17419` and the workflow silently stops producing output.

Measured in prod (read-only) on the offending workflow's conversation:
```
doc = 16.0 MB | messages = 1,703 | avg 9.6 KB/msg | max 1.1 MB/msg
```
Pure accumulation — bounding the count fixes it.

## Fix

- `update_messages(..., max_messages=None)` — when set, the `$push` uses `$slice: -max_messages`, keeping only the most recent N messages atomically.
- The per-workflow conversation path (`add_workflow_execution_messages`) passes `WORKFLOW_CONVERSATION_MAX_MESSAGES = 50` — plenty of recent thread context, ~2 orders of magnitude under 16MB at the observed avg message size.
- **Regular chat conversations pass no cap and are completely unchanged.**

## Verification (local Mongo)

```
push 60 w/ cap=50  -> len=50  (m10..m59)   # oldest pruned
push +20 w/ cap=50 -> len=50  (m30..m79)   # window slides, stays bounded
push +10 NO cap    -> len=60                # chat path still grows (unchanged)
PRUNE_VERIFIED: True
```
`ruff` + `ruff format` clean; `mypy` clean; `pytest tests/unit -k "conversation or workflow"` → **603 passed**.

## Notes
- Count-based `$slice` is atomic and sufficient here (avg msg ~9.6 KB). Pathologically large *single* messages (>16MB) are a separate concern not addressed here.
- Existing already-oversized conversations won't shrink until the next append trims them; a one-off backfill could prune them immediately if desired.
- Part of the workflow-reliability set with #844 (rate limit) and #846 (trigger matching).

### DIFF
--- a/apps/api/app/config/rate_limits.py
+++ b/apps/api/app/config/rate_limits.py
@@ -154,8 +154,10 @@ FEATURE_LIMITS: dict[str, TieredRateLimits] = {
         ),
     ),
     "trigger_workflow_executions": TieredRateLimits(
-        free=RateLimitConfig(day=5, month=20),  # Keep restrictive
-        pro=RateLimitConfig(day=150, month=4500),  # +50% (100→150, 3000→4500)
+        free=RateLimitConfig(day=5, month=20),  # Keep restrictive (trial)
+        # Account-level triggers (e.g. Gmail) fire one execution per event, so Pro
+        # needs real headroom; the cap stays as a cost guard, not a daily blocker.
+        pro=RateLimitConfig(day=1500, month=45000),
         info=FeatureInfo(
             title="Trigger Workflow Executions",
             description="Automated workflow executions triggered by integrations",
--- a/apps/api/app/services/conversation_service.py
+++ b/apps/api/app/services/conversation_service.py
@@ -1,5 +1,6 @@
 import asyncio
 from datetime import UTC, datetime
+from typing import Any
 from uuid import uuid4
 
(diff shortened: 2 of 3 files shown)


aryanranderiya:
Folded in two more changes per request:

**Raise Pro `trigger_workflow_executions` limits (10x):** `150 → 1500/day`, `4500 → 45000/month`. Account-level triggers (Gmail) fire one execution per event, so Pro needs real headroom rather than a daily blocker. Free stays restrictive (5/20, trial), and the cap remains as a cost guard. Raising the numbers (vs zeroing the cap) keeps the rate-limit invariants intact (`pro ≥ free`, `month ≥ day`) — no test changes; 657 unit tests pass. Numbers are tunable to your unit economics.

**Docstring cleanup:** trimmed the verbose `update_messages` docstring + per-workflow comments to concise one-liners per the apps/api docstring guidance.

PR #849, opened 2026-07-02
feat(workflows): Inbox Triage -> daily digest; filter Auto-Draft trigger; reseed drift + orphans

aryanranderiya:
## What

One-shot, idempotent, **dry-run-first** script (`apps/api/scripts/reseed_system_triggers.py`) that consolidates the Composio trigger mess found during the trigger-pipeline investigation (companion to #846):

1. **De-aggress Gmail system workflows.** "Inbox Triage" / "Auto-Draft Replies" fire one full agent run per inbound email. Their Composio poll trigger is re-created with a `labelIds` filter (default `IMPORTANT`, `--label` to override) so Composio only fires for mail Gmail marks important. The filter lives entirely in Composio's trigger config — **no workflow schema, API, or modal changes**.
2. **Fix reconnect drift.** Workflows whose stored `composio_trigger_ids` are no longer active on Composio are re-registered and repointed at the live nano-id (Composio upserts, so this is idempotent). `calendar_ids=["all"]` expands exactly like `CalendarTriggerHandler._fetch_user_calendars` (incl. its primary fallback).
3. **Delete orphans.** Active Composio triggers referenced by no workflow (deletes/reconnects left them firing dead webhooks) are removed — recomputed *after* steps 1–2 so re-adopted ids are never touched.

## Safety

- **Dry-run by default.** Writes require **both** `--apply` and `--yes`; `--apply` alone aborts.
- Skips any user without an ACTIVE Composio connection for the relevant toolkit (nothing to register against).
- Idempotent: re-running upserts to the same trigger ids; `unchanged` is reported.

## Prod dry-run results (read-only, run today)

```
gmail:    updated=112  unchanged=0  skipped=253   (253 = users whose Gmail connection is EXPIRED/REVOKED)
calendar: fixed=2      skipped=33                 (33 = users who disconnected Google Calendar)
orphans:  active=791   referenced=394  would delete 397
```

Notable: **~70% of Gmail system workflows belong to users with no active Gmail connection** — they can't fire regardless of any fix and are correctly skipped (they'd need a reconnect-time re-registration hook, tracked separately).

## Run

```bash
# dry-run (safe, read-only)
cd apps/api && uv run python scripts/reseed_system_triggers.py

# apply (after reviewing dry-run; intended AFTER #846 is deployed)
cd apps/api && uv run python scripts/reseed_system_triggers.py --apply --yes
```

Sequencing note: apply is safe before or after #846 deploys (new ids only *match* once #846 is live; the label filter takes effect at Composio immediately), but the natural order is #846 first.


aryanranderiya:
**Updated per review discussion — Inbox Triage is now converted to a daily scheduled digest instead of a filtered poll trigger.**

The two Gmail system workflows have different natures, so they're reshaped differently:

| Workflow | Nature | New shape |
|---|---|---|
| **Inbox Triage** | digest job (its value is a batched summary) | **daily SCHEDULE** at 08:00 user-local, prompt rewritten to fetch the last 24h itself. Exits Composio trigger machinery entirely — runs on the scheduler path that already works reliably in prod. |
| **Auto-Draft Replies** | timeliness is the feature | stays **event-driven** with the `labelIds: IMPORTANT` filter (unchanged from previous revision) |

Code changes beyond the script: `definitions/gmail.py` (scheduled Inbox Triage definition — new users get it at provisioning) and `provisioner.py` (stamps the user's profile timezone onto scheduled system configs so the cron fires at local time, not UTC).

**Safety additions:**
- Script section 1a reads the **current** definition and **aborts if the build's definition isn't the scheduled version** — verified live: running today's script against the current prod image prints the abort instead of converting to a wrong shape. **Apply must happen after this PR deploys.**
- Writes still require both `--apply` and `--yes`.

**Latest prod dry-run (read-only):** 1a `ABORT (old image — guard working)`, 1b drafts `updated=56 skipped=126`, calendar `fixed=2 skipped=33`, orphans `397`.

**Verification:** 23 system-workflow/provisioner unit tests pass; `next_run` verified for Asia/Calcutta (02:30 UTC = 08:00 IST), America/New_York, UTC; ruff + mypy clean.

aryanranderiya:
**Two more changes, prompted by the reconnect question:**

**1. Root fix for the drift: trigger resync on (re)connect.** Confirmed mechanism: reconnecting creates a *new* Composio connected account, and `handle_oauth_connection` only re-subscribed account-level `auto_activate` triggers — per-workflow triggers were never re-registered, so every trigger-id-matched workflow kept pointing at the old account's dead ids and went permanently silent. All 171 stale workflows trace to this. New `TriggerService.resync_user_workflow_triggers` runs as a background task on every (re)connect: re-registers the user's activated integration workflows against the current account, repoints `composio_trigger_ids`, unregisters the stranded old ids. **Without this, the reseed script would need re-running after every wave of reconnects.**

**2. Orphan deletion made provably safe.** "Unreferenced by `composio_trigger_ids`" ≠ unused: the account-level GMAIL trigger feeds `gmail_new_message` workflows (matched by `user_id`, no stored ids), and `auto_activate` subscriptions (calendar EVENT_CREATED) are recreated on every connect by design. Deletion is now restricted to provably per-workflow triggers. Prod dry-run validates the hazard was real:

```
unreferenced=397 -> deletable=231
kept: gmail-account-level-in-use=29   <- previous policy would have deleted these and
                                         broken the only currently-WORKING gmail workflows
kept: auto-activated-account-level=137
```

Latest full prod dry-run (read-only): 1a `ABORT` (old image — version guard), 1b drafts `56/126-skipped`, calendar `2/33-skipped`, orphans `231 deletable`. 994 unit tests pass, ruff + mypy clean. No writes have been performed at any point.

aryanranderiya:
**Final safety refinement: nothing is deleted anymore.** All Composio trigger removals (orphan pass + old poll triggers replaced in 1a/1b) now use `triggers.disable()` instead of `delete` — same effect (stops polling + dead webhooks), but fully reversible via `triggers.enable(trigger_id)`. On apply, the disabled ids are written to a `system_migration_audit` doc so rollback is a one-liner over the exact set. Re-adoption remains the first resort: Composio's upsert means re-registration repoints workflows at an existing matching trigger before the orphan pass ever sees it. Prod dry-run unchanged: `to_disable=231`, kept `29` gmail account-level in-use + `137` auto-activated. Still zero writes performed.

aryanranderiya:
**Disable-set corrected after a per-trigger forensic audit — 231 → 74.**

Reviewer skepticism was right: most of the "orphans" were NOT from deleted workflows. Audited each trigger's own config + its owner's workflows:

| Class | Count | Origin | Disposition |
|---|---|---|---|
| GMAIL `interval=1` | **157** | connect-time account subscription (`handle_subscribe_trigger` on every OAuth connect) — unreferenced **by design**, load-bearing for any future `gmail_new_message` workflow | **KEEP** (was wrongly in the disable set) |
| GMAIL `interval=15/30/2` | 24 | per-workflow poll leftovers — 95 owners have an activated poll workflow pointing at a *different* id (reconnect drift), 82 have none | disable |
| STARTING_SOON | 50 | only ever created per-workflow; 49 owners have **no** starting-soon workflow at all | disable |

Nothing in the disable set can match any workflow (matching is by stored id — none stores these — or by `user_id` for `gmail_new_message`, whose users are excluded twice over). Prod dry-run: `to_disable=74 | kept: in-use=29, connect-time=157, auto-activated=137`. Still dry-run only; still reversible + audited.

aryanranderiya:
**Verification sweep of the remaining assumptions:**

**1. Converted Inbox Triage actually fires — VERIFIED (local, real code).** Seeded a poll-shaped system workflow, ran the script's real `convert_triage_to_daily`, backdated `scheduled_at`, then ran the actual recovery scan (`workflow_scheduler.scan_and_schedule_pending_tasks`): the `execute_workflow_by_id` ARQ job was enqueued with `trigger_type=schedule` (claim/re-arm path), and a deactivated workflow was correctly NOT enqueued. The full chain conversion → scan → enqueue works.

**2. `list_active` excludes disabled triggers — VERIFIED (prod, read-only).** Default listing = 791; `show_disabled=True` = 1,165 (delta 374, all `disabled=True`). Disable semantics + script idempotency confirmed. (Side-finding: 374 triggers are already disabled on Composio from prior cleanup.)

**3. `labelIds=IMPORTANT` filtering — documented, NOT live-verified.** No trigger in prod has ever used a non-INBOX filter (381 INBOX / 31 unset), so there is no read-only evidence of the filter in action. Composio's trigger schema explicitly documents labelIds as message filtering ('Filter messages by a single label ID… IMPORTANT…'), but the first real proof will be either a one-off test trigger on a consenting account or the first post-apply important-vs-newsletter email. Flagging honestly rather than claiming verified.

aryanranderiya:
**New root-cause evidence: Composio auto-disables triggers when a connected account expires.**

While attempting a live labelIds-filter test, discovered the test user's Gmail account is `EXPIRED` (410 on trigger create). Cross-referencing: their two Gmail triggers show `disabled_at = 2026-06-30T21:54:52.685Z` — both at the same millisecond the account expired. Zooming out, the `show_disabled` listing shows **272 of the 374 disabled triggers were disabled in one 2-hour window on 2026-05-23** — the exact window the same user's older accounts went EXPIRED/REVOKED. So the full lifecycle of the drift is now evidenced end-to-end:

```
token expiry (mass event May 23; rolling since)
  -> Composio auto-disables the account's triggers
  -> user reconnects -> NEW connected account, NEW trigger ids
  -> nothing re-registered per-workflow triggers  <- fixed by resync_user_workflow_triggers in this PR
  -> workflows point at old ids forever (stale) + new/old triggers unreferenced (orphans)
```

This validates the design: the ConnectionGuard's ACTIVE check correctly skips expired users, and the resync-on-reconnect hook is exactly the self-heal those users need.

**labelIds filter live test: blocked** — the only consented account (the requester's) is expired, and re-auth is interactive by nature. Test scripts are ready to run unattended the moment the account is reconnected; until then the filter remains documented-but-not-live-verified (worst case if it under-filters: behavior equals today's).

apps/api/scripts/reseed_system_triggers.py
```diff
@@ -0,0 +1,530 @@
...
+from datetime import UTC, datetime
+import os
+from pathlib import Path
+import sys
+
+sys.path.insert(0, str(Path(__file__).parent.parent))
+
+# In the production container, Infisical credentials live in docker secrets and
+# are exported by the entrypoint — exec'd scripts must load them themselves.
+_SECRET_ENV = {
```
aryanranderiya: remove this script. we only want it locally

apps/api/app/services/workflow/trigger_service.py
```diff
@@ -184,3 +185,56 @@ async def unregister_triggers(
         except Exception as e:
             log.error(f"{LogTag.WORKFLOW} Error unregistering triggers: {e}")
             return False
+
+    @staticmethod
+    async def resync_user_workflow_triggers(user_id: str, trigger_names: list[str]) -> None:
```
aryanranderiya, line 190: have we tested this?

apps/api/scripts/reseed_system_triggers.py
aryanranderiya: Done — removed from the repo in 06dfcbaa8. Kept locally at ~/Projects/GAIA/gaia/.agents/scripts/reseed_system_triggers.py (gitignored); it runs by piping into the prod backend container, so it never needed to ship in the image.

apps/api/app/services/workflow/trigger_service.py
aryanranderiya: It wasn't — fair catch. Added tests/unit/services/test_trigger_resync.py in 06dfcbaa8 with 5 cases: repoints ids + unregisters old on success, no-op on empty trigger names, skips account-level registrations that return no ids, skips unchanged ids, and one workflow's failure doesn't block the rest. All pass. Not live-tested against a real reconnect yet — the one consented account (Aryan's Gmail) is currently EXPIRED; the moment it's reconnected this hook fires for real and the resync log line will confirm it.

PR #878, opened 2026-07-21
fix(api): drop removed ast aliases so config schema dump works on Python 3.12+

gusmclennan:
## Problem

`gaia setup` and `gaia init` fail during environment configuration on Python 3.12 and newer:

```
File "apps/api/scripts/dump_config_schema.py", line 121, in visit_ClassDef
    elif isinstance(body_item.value, ast.Str):
                                     ^^^^^^^
AttributeError: module 'ast' has no attribute 'Str'
```

`dump_config_schema.py` references `ast.Str` (8×) and `ast.NameConstant` (2×). Both were deprecated in Python 3.8 and **removed in Python 3.12**.

The CLI shells out to this script from `packages/cli/src/lib/env-parser.ts` (`parseSettings`) to build the env-var prompt list. When it throws, `env-setup.ts` catches it and reports:

> Manual environment setup requires Python to parse config schema.
> For self-host mode, we recommend using Infisical for secret management.
> Alternatively, install Python 3.11+ and try again.

That message misdirects — the user *does* have Python 3.11+, they have something newer. The practical effect is that manual self-host setup is broken for anyone on a current Python, with no obvious path forward short of downgrading or adopting Infisical.

Reproduced on macOS with Python 3.14.6.

## Fix

Remove the dead legacy branches.

Every one of them was already unreachable: each `elif isinstance(..., ast.Str)` sits behind an `isinstance(..., ast.Constant)` check that matches first, and the two `ast.NameConstant` checks are `or`-ed after the same. `ast.Constant` has represented str/bytes/numbers and `True`/`False`/`None` since Python 3.8, so nothing in the repo's supported range ever reached them. They only caused trouble because `isinstance()` still evaluates the attribute lookup at runtime.

Net: 3 insertions, 17 deletions, one file. No behaviour change.

## Verification

- Script exits 0 on Python 3.14.6 with empty stderr (previously `AttributeError`)
- Emitted JSON is **byte-identical** before and after the change — compared against a baseline captured by monkey-patching the removed aliases back in, across all 21 categories
- Repo-wide grep confirms no other Python file references a removed `ast` alias


Dhruv-Maradiya:
Hey @gusmclennan thank you for your contribution it means a lot to us

Dhruv-Maradiya (approved):

PR #879, opened 2026-07-21
feat(api): support E2B_DOMAIN for the EU cluster migration

Dhruv-Maradiya:
## What

E2B was migrated from the US region to the EU cluster (`e2b-juliett.dev`). This wires `E2B_DOMAIN` through the API.

## How it works

No call-site plumbing was needed. The e2b SDK resolves its cluster from the **process env**, not from a constructor arg:

```python
# e2b/connection_config.py
@staticmethod
def _domain():
    return os.getenv("E2B_DOMAIN") or "e2b.app"
```

Verified live:

```
E2B_DOMAIN=e2b-juliett.dev  ->  domain = e2b-juliett.dev, api_url = https://api.e2b-juliett.dev
```

This is the same mechanism `E2B_API_KEY` already uses — the code never passes `api_key=` either. `AsyncSandbox.create`, `AsyncSandbox.connect` and `Template.build` all build a `ConnectionConfig` with no explicit domain, so all three follow the env.

`settings.py` runs `load_dotenv()` and `inject_infisical_secrets()` writes every secret into `os.environ` *before* Pydantic builds the settings object, so an Infisical value reaches the SDK. `settings = get_settings()` at module scope means the API and the ARQ worker share that path identically.

## Changes

| File | Change |
|---|---|
| `app/config/settings.py` | `E2B_DOMAIN: str` (required in prod), `str \| None = None` in dev — matching the `E2B_API_KEY` / `E2B_TEMPLATE_ID` pattern |
| `app/services/sandbox/lifecycle.py` | sandbox-create log includes `domain=`, so a wrong cluster is visible instead of surfacing as a confusing "template not found" |
| `scripts/build_e2b_template.py` | requires `E2B_DOMAIN` and prints the cluster it is building against |
| `.env.example` | documents `E2B_DOMAIN` |

**No default on purpose.** Templates and API keys are per-cluster, so falling back to the SDK's `e2b.app` would silently point EU credentials at the US cluster. Same reasoning in the build script, which now exits rather than building against the wrong cluster.

## Deploy steps

`E2B_DOMAIN` must be in Infisical **before** this merges. The field is required in `ProductionSettings`, and a missing required key makes the API **refuse to boot** — `from_env()` logs `[STARTUP] Error creating settings: ... E2B_DOMAIN Field required`, then its fallback loop skips the field (a required field's `default` is `PydanticUndefined`, so the `is not None` guard `continue`s) and the retry re-raises. Add it to every env slug (it is currently absent from `development`). Same contract as every other required prod key.

The template must be rebuilt on the EU cluster and `E2B_TEMPLATE_ID` updated (the current `0tnfu1c42a4xd07bsg5p` is a US-cluster template):

```bash
cd apps/api
E2B_DOMAIN=e2b-juliett.dev E2B_API_KEY=<EU-cluster-key> \
  uv run python scripts/build_e2b_template.py
```

Then re-run the hardening check (`scripts/verify_sandbox_hardening.sh`), since sudo-strip / hidepid are template-baked.

## Scope check

Grepped the monorepo: only `apps/api` touches e2b. No JS/TS package depends on the SDK, voice-agent and bots have no `E2B_` references, both compose files already pass `env_file: apps/api/.env`, and there are no `E2B_` references in CI. One place to set it: Infisical (plus `.env` locally).

## Verification

- `ProductionSettings.E2B_DOMAIN` -> `required=True`; `DevelopmentSettings` -> `required=False, default=None` (introspected `model_fields`)
- Dev load with the var unset -> `None`; set -> `'e2b-juliett.dev'`
- Build script without `E2B_DOMAIN` -> `E2B_DOMAIN is not set in the environment`, exit 1
- `ruff check` + `ruff format --check` clean, `mypy` clean on both touched modules, `tests/unit/sandbox` 28 passed

## Blank / padded values (added after review)

`E2B_DOMAIN` being a plain `str` was not enough — pydantic accepts `""`, and the SDK reads the env var verbatim:

```
env=''                  -> api_url='https://api.e2b.app'          # silently the US cluster
env='   '               -> api_url='https://api.   '              # malformed
env=' e2b-juliett.dev ' -> api_url='https://api. e2b-juliett.dev ' # malformed
```

The first case is exactly the silent wrong-cluster failure this PR exists to prevent, and it produced no startup signal. A validator on `CommonSettings` (`check_fields=False`, since the field is declared per-environment in the subclasses) now rejects blank and surrounding-whitespace values at boot. It rejects rather than strips, because a stripped *settings* value would not change what the SDK reads out of `os.environ`.

Verified against the real settings classes:

| `E2B_DOMAIN` | result |
|---|---|
| unset (dev) | boots, `None` |
| `''` | refuses to boot |
| `'   '` | refuses to boot |
| `' e2b-juliett.dev '` | refuses to boot |
| `'e2b-juliett.dev'` | boots |

## Known gap (not addressed here)

The `E2B Code Execution` group in `settings_validator.py` registers only `E2B_API_KEY` — neither `E2B_TEMPLATE_ID` nor `E2B_DOMAIN` gets the actionable "missing configuration" startup warning. Pre-existing; worth a follow-up.

### DIFF
--- a/apps/api/.env.example
+++ b/apps/api/.env.example
@@ -107,9 +107,9 @@ ENABLE_OPENUI=false
 # tools registered-but-disabled — they return a clear "sandbox not configured"
 # error at call time.
 
-# E2B cluster domain. Read directly from the env by the e2b SDK. Defaults to
-# `e2b.app` (US); GAIA runs on the EU cluster. The API key and the template ID
-# are per-cluster — all three must belong to the same cluster.
+# E2B cluster domain. Read directly from the env by the e2b SDK. GAIA runs on
+# the EU cluster; the API key and the template ID are per-cluster, so all three
+# must belong to the same cluster.
 # E2B_DOMAIN=e2b-juliett.dev
 
 # E2B sandbox template ID — produced by (with E2B_API_KEY + E2B_DOMAIN exported):
--- a/apps/api/.env.example
+++ b/apps/api/.env.example
@@ -107,7 +107,12 @@ ENABLE_OPENUI=false
 # tools registered-but-disabled — they return a clear "sandbox not configured"
 # error at call time.
 
-# E2B sandbox template ID — produced by:
+# E2B cluster domain. Read directly from the env by the e2b SDK. Defaults to
+# `e2b.app` (US); GAIA runs on the EU cluster. The API key and the template ID
+# are per-cluster — all three must belong to the same cluster.
+# E2B_DOMAIN=e2b-juliett.dev
+
+# E2B sandbox template ID — produced by (with E2B_API_KEY + E2B_DOMAIN exported):
 #   uv run python apps/api/scripts/build_e2b_template.py
 # E2B_TEMPLATE_ID=
 
--- a/apps/api/app/config/settings.py
+++ b/apps/api/app/config/settings.py
@@ -155,6 +155,16 @@ class CommonSettings(BaseAppSettings):
     # - Used for discovering and installing skills from GitHub
     GITHUB_TOKEN: str | None = None
 
+    # check_fields=False: E2B_DOMAIN is declared per-environment in the subclasses.
+    # Rejected rather than stripped because the e2b SDK reads os.environ verbatim —
+    # "" there silently falls back to the US cluster, padding yields a broken URL.
+    @field_validator("E2B_DOMAIN", mode="after", check_fields=False)
+    @classmethod
+    def _reject_unusable_e2b_domain(cls, v: str | None) -> str | None:
+        if v is not None and (not v or v != v.strip()):
+            raise ValueError("E2B_DOMAIN must be non-empty and free of surrounding whitespace")
+        return v
+
     # ----------------------------------------------
     # Computed Properties
     # ----------------------------------------------
--- a/apps/api/app/config/settings.py
+++ b/apps/api/app/config/settings.py
@@ -295,6 +295,12 @@ class ProductionSettings(CommonSettings):
     # ----------------------------------------------
     E2B_API_KEY: str
     E2B_TEMPLATE_ID: str  # gaia-coder template ID (run scripts/build_e2b_template.py)
+    # E2B cluster the SDK talks to. Like E2B_API_KEY, the SDK reads this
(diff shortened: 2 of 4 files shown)


Dhruv-Maradiya:
> Reject blank `E2B_DOMAIN` values.

Good catch — verified and fixed in 0f6bcd1. The empty-string case is worse than described: it doesn't just break sandbox creation, it **silently routes to the wrong cluster**, which is the exact failure this PR exists to prevent.

What the SDK actually does with each value (run against `e2b.connection_config.ConnectionConfig`):

```
env=''                  -> api_url='https://api.e2b.app'           # silently the US cluster
env='   '               -> api_url='https://api.   '               # malformed
env=' e2b-juliett.dev ' -> api_url='https://api. e2b-juliett.dev '  # malformed
```

So I widened the check past non-blank to also reject surrounding whitespace — a trailing space copy-pasted into Infisical is just as broken, and your suggested check wouldn't have caught it.

Two implementation notes:

**It rejects rather than strips.** The SDK reads `E2B_DOMAIN` from `os.environ` directly (`ConnectionConfig._domain()`); it is never passed at a call site. A stripped *settings* value would not change what the SDK reads, so normalizing here would look like a fix without being one. Refusing to boot is the honest option.

**Placed on `CommonSettings` with `check_fields=False`,** since the field is declared per-environment (`str` in prod, `str | None` in dev) and duplicating the validator in both subclasses would be the same code twice.

Verified against the real settings classes rather than a stub:

| `E2B_DOMAIN` | result |
|---|---|
| unset (dev) | boots, `None` |
| `''` | refuses to boot |
| `'   '` | refuses to boot |
| `' e2b-juliett.dev '` | refuses to boot |
| `'e2b-juliett.dev'` | boots |

`ruff check` / `format` clean, `mypy` clean. `tests/unit` shows 8 pre-existing failures in `test_auth_middleware.py` — identical with and without this change, and unrelated to E2B.

I also corrected the PR description: it previously claimed a missing required key coerces to `""`. It does not — `from_env()`'s fallback skips required fields (their `default` is `PydanticUndefined`, not `None`), so the retry re-raises and the API refuses to boot.

PR #891, opened 2026-08-05
feat(agents): env-configurable custom dev LLM provider for cheap bulk testing

aryanranderiya:
## What & why

Bulk dev/test chat traffic currently burns real OpenRouter credits. This adds a dev-only **`custom` LLM provider slot**: any OpenRouter/OpenAI-compatible endpoint, configured entirely by env —

```bash
DEV_LLM_BASE_URL=https://inference-api.nousresearch.com/v1   # e.g. Nous Research's 90%-off DeepSeek lane
DEV_LLM_API_KEY=<key>
DEV_LLM_MODEL=deepseek/deepseek-v4-flash-0731                # $0.01/M in, $0.02/M out on Nous
DEV_DEFAULT_MODEL=custom                                     # optional: route every dev request to it by default
```

The slot is deliberately generic (not Nous-specific): the same three vars point at OpenCode Zen (`https://opencode.ai/zen/v1`, `deepseek-v4-flash-free`) or any other cheap OpenRouter-compatible lane with zero code changes. The dev menu also gains **DeepSeek V4 Flash** on OpenRouter (same pinned `-0731` snapshot, $0.09/$0.18 there), so the two routes can be A/B'd on the identical model.

**Production is untouched**: the provider is registered only when `ENV=development` (`register_llm_providers`), so `DEV_LLM_*` present in a prod environment cannot route traffic.

## How it works

- `app/agents/llm/client.py` — `init_custom_llm()` registers a `ChatOpenRouter` with a `base_url` override (the Nous/Zen APIs are OpenRouter-wire-compatible, so reasoning parsing, retries, and the fallback chain behave identically to the OpenRouter client). It joins the existing `configurable_alternatives` chain as provider `custom`, follows the `GAIA_SIM_MODE` stub pattern, and shares the OpenRouter client's configurable-field ids.
- `app/agents/llm/plan_model.py` — requests that don't pick a model (`use_default_models`) now fall back to `DEV_DEFAULT_MODEL` for both comms and executor, so bots/scripts/plain requests route there too; an explicit chat-header selection still wins. A typo'd value warns and keeps the plan model. Menu entries without a pinned model (`custom`) route by provider and clear stale model pins so the client's `DEV_LLM_MODEL` default serves the request.
- `app/constants/llm.py` — `DEV_MODEL_OPTIONS` gains `deepseek-v4-flash` (OpenRouter) and `custom`; `app/config/settings.py` gains the four `DEV_LLM_*`/`DEV_DEFAULT_MODEL` fields on `CommonSettings` (optional, dev-honored only).
- `apps/web .../devModels.ts` — matching **DeepSeek V4 Flash** and **Custom (env)** picker entries.
- `.env.example` — documents the LLM provider vars (previously absent entirely). Note: comments sit **above** the empty `GOOGLE_API_KEY=`/`OPENROUTER_API_KEY=` lines because python-dotenv parses an inline comment after an *empty* value as the value itself — found live when `Bearer # OpenRouter — paid-plan + dev-menu models` became an Authorization header and broke every chat turn.

## Before / After

| Surface | Before | After |
|---|---|---|
| Dev model picker |  |  |
| Custom (env) pinned for comms + executor | — |  |
| Chat turn answered through the custom provider | — |  |

## Verification

- Full stack booted in the sandbox (infra in Docker, native API + web, dev auth bypass, seeded dev user per `driving-gaia`). The repo's OpenRouter-wire `llm-stub` played the role of the external endpoint via `DEV_LLM_BASE_URL=http://localhost:9797/api/v1`.
- **Custom provider, explicit pin**: `POST /chat-stream` with `use_default_models: false, comms_model: "custom"` → real (non-sim) graph run made a live HTTP call to the env-configured endpoint and streamed the reply; stub log shows the request with the full comms toolset bound.
- **Custom provider, env default**: same request with *no* model fields under `DEV_DEFAULT_MODEL=custom` → routed to the endpoint, winning over the free-plan Gemini pin as designed.
- **From the browser**: drove `/c` with Playwright as the seeded dev user — selected Custom (env) for both roles in the chat-header picker, sent a message, reply rendered in the chat UI (screenshot above).
- Endpoint compatibility verified against the live `/v1/models` of both Nous (`deepseek/deepseek-v4-flash-0731`, 1M ctx, tools + reasoning params) and OpenCode Zen (`deepseek-v4-flash-free`) — both are OpenRouter-schema services.
- Local gates green: `nx type-check api`, `nx lint api`, `nx run web:type-check`, `nx lint web`; `test_llm_client.py` + `test_llm_fallback_chain.py` (69 passed, assertions updated for the new provider).
- Run modes used: `--sim`-equivalent for stack bring-up checks, then non-sim with `DEV_LLM_*` for the custom-provider path (sim mode by design replaces every factory with the stub, so it cannot exercise this provider).

## Not verified

- No real Nous/OpenCode API calls (no credentials in this sandbox) — the wire contract was exercised against the repo's OpenRouter-compatible stub and both providers' live model catalogs, not a paid completion.
- Compaction/summarization thresholds remain denominated in the default model's window for dev-menu overrides (pre-existing, documented in `constants/llm.py`).

https://claude.ai/code/session_015SHeRbCJFybQHTSqhK5KQZ

---
_Generated by [Claude Code](https://claude.ai/code/session_015SHeRbCJFybQHTSqhK5KQZ)_

### DIFF
--- a/apps/api/.env.example
+++ b/apps/api/.env.example
@@ -71,6 +71,25 @@ KAPSO_WEBHOOK_SECRET=
 WHATSAPP_PHONE_NUMBER=
 
 # =============================
+# LLM Providers
+# =============================
+# Gemini — default model + auxiliary LLM tasks
+GOOGLE_API_KEY=
+# OpenRouter — paid-plan + dev-menu models
+OPENROUTER_API_KEY=
+
+# Dev-only (honored only when ENV=development):
+# Custom OpenRouter/OpenAI-compatible endpoint for cheap bulk test usage — all three must
+# be set. Example: Nous Research's discounted DeepSeek lane (https://portal.nousresearch.com).
+# DEV_LLM_BASE_URL=https://inference-api.nousresearch.com/v1
+# DEV_LLM_API_KEY=
+# DEV_LLM_MODEL=deepseek/deepseek-v4-flash-0731
+# Default model for every dev request that doesn't pick one in the chat-header selector —
+# any DEV_MODEL_OPTIONS key from app/constants/llm.py ("custom" = the endpoint above).
+# An explicit selector choice wins.
+# DEV_DEFAULT_MODEL=custom
+
+# =============================
 # GitHub Integration (optional)
 # =============================
 # Get a token at https://github.com/settings/tokens (no scopes needed)
--- a/apps/api/app/agents/llm/client.py
+++ b/apps/api/app/agents/llm/client.py
@@ -27,6 +27,7 @@ from app.constants.llm import (
     DEFAULT_LLM_TEMPERATURE,
     DEFAULT_MAX_TOKENS,
     DEFAULT_MODEL_NAME,
+    DEV_LLM_MAX_OUTPUT_TOKENS,
     LLM_INVOKE_TIMEOUT_SECONDS,
     LLM_RETRY_MAX_ATTEMPTS,
     OPENROUTER_APP_CATEGORIES,
@@ -74,10 +75,14 @@ def is_default_model_config(configurable: Mapping[str, Any]) -> bool:
 PROVIDER_MODELS = {
     "gemini": DEFAULT_GEMINI_MODEL_NAME,
     "openrouter": DEFAULT_GROK_MODEL_NAME,
+    # The env-defined custom dev endpoint; empty when unset — the provider is
+    # only registered in development with all DEV_LLM_* settings present.
+    "custom": settings.DEV_LLM_MODEL or "",
 }
 PROVIDER_PRIORITY = {
     1: "gemini",
     2: "openrouter",
+    3: "custom",
 }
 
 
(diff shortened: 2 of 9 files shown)


aryanranderiya:
## Live verification — custom lane works end-to-end ✅

Tested against a full local stack (`mise dev --agent`, this PR's diff applied, `DEV_LLM_*` pointed at Nous):

**Direct API (curl):** `deepseek/deepseek-v4-flash-0731` completes and streams correctly on `https://inference-api.nousresearch.com/v1` (OpenAI-wire SSE, served via Novita). Note: the model id only resolves **with a valid key** — unauthenticated probes return "Unknown model", which looks like a wrong id but is just key-scoping.

**Through the product (browser-driven):** dev picker → both COMMS and EXECUTOR pinned to **Custom (env)** → sent a research prompt → comms replied, handed off to the executor, executor ran real web-search tools, full answer streamed and persisted to Mongo. No LLM fallback fired — the whole turn ran on the custom provider.

| Picker | Chat on the custom lane |
|---|---|
| picker | chat |

**Behavioral note:** DeepSeek V4 Flash emits `reasoning` tokens by default (the `reasoning: False` option entry doesn't suppress them). They count as output tokens — still ~free at this lane's pricing, but budget-relevant if anyone reuses the slot for a pricier endpoint.

**Also on this branch:** `5b5930f` fixes `scripts/payment_setup.py`, which #880's `PlanDB` → `PlanDocument` rename had broken (found while seeding a fresh dev Mongo for this test).

**Review verdict:** architecture clean — dev-only registration with a prod-gating unit test, env-example accurate, no debt smells. Two non-blocking notes: `custom` joins the dev fallback chain (priority 3), so a Gemini+OpenRouter outage in dev silently lands on the cheap model; and `PROVIDER_MODELS["custom"]` is frozen at import, so changing `DEV_LLM_MODEL` needs an API restart. Ready to merge.

The `assets/pr891-verification` branch only carries these two screenshots — safe to delete after merge (the images above will stop rendering).

aryanranderiya:
**Addendum — routing proven with a negative control.**

To rule out "the turn silently ran on Gemini while the picker said custom": restarted the stack with `DEV_LLM_API_KEY` intentionally broken and sent the same explicitly-pinned request. The stream immediately surfaced

```
data: {"model_fallback": {"model": "gemini-3.1-flash-lite"}}
```

— i.e. the primary attempt goes to the custom endpoint, fails auth, and the fallback chain reports itself. With the valid key, identical turns emit **zero** `model_fallback` events. So valid-key turns are conclusively served by the Nous lane, and `model_fallback` in the SSE stream is the observable signal to watch when smoke-testing this slot.

(Side note for future testers: the endpoint publishes AAAA records, so the connection rides IPv6 — IPv4-only network snooping will miss it.)

PR #902, opened 2026-08-07
test(api): test-infrastructure overhaul — canonical tiers, hermetic default, CI gates, gap-fill

aryanranderiya:
## Test Infrastructure Overhaul

Full plan (research base, decisions, taxonomy, done-when): https://gist.github.com/aryanranderiya/db9e49da76ba314267de6509e3eaf628

### What this PR does
- **Restructure**: canonical tier layout — `unit/`, `integration/`, `integration/real/` (+ `real/memory/`), `contracts/`, `e2e/`, `stress/`, `meta/`. Duplicates/dead files deleted.
- **Hermetic by default**: root conftest fences every test (blanked creds, deterministic env); `USE_REAL_SERVICES` defaults off locally; `nx test api` = fast offline suite.
- **New tiers**: meta-tests (import fence + architecture invariants), stress tier (ARQ idempotency, webhook replay, SSE resume, claim race), Schemathesis contract gate (real server, HTTP).
- **Harness upgrade**: input-recording fake LLM + call-all-tools model + prompt-contract tests.
- **Gap-fill**: ~60 new test files (services, agents, tools, endpoints, worker crons) + 22 property tests.
- **Defects**: unfalsifiable asserts rewritten, real sleeps → deterministic, pre-existing `test_agent_core` failure fixed, paywall/rate-limiter un-blinded (429 tests), flaky debounce test fixed deterministically.
- **CI**: JUnit artifacts + failure summary, `--durations`, diff-cover gate, **regression-proof job** (changed tests must fail on base), semgrep lane (informational), flake-gate script.
- **Docs**: `tests/CLAUDE.md`, scaffolds, root AGENTS.md rule flipped to tests-first-class.

### Verification (locally reproduced CI-equivalent, all green)
- Full tree with live services + coverage: **11,192 passed, 0 failed** (after local-port fix — CI maps Chroma on 8000, local compose on 8080; Postgres creds note in `integration/real/README.md`)
- gaia-shared: 200 · tools/lints: 66 · vitest (web/desktop/bots/shared/cli): 5 projects green
- Ruff + mypy clean

### Honest caveats / follow-ups
- **Schemathesis**: scoped gate green; full-schema fuzz documented real contract drift (undocumented 429 plan-gates, 405-vs-422 route shadowing, 403/307/410/404) + a server-wedge under pathological fuzz input — needs the API-hardening follow-up (see `tests/integration/real/test_schemathesis.py` docstring)
- **Warnings-as-errors deferred**: suite emits ~1.9k third-party pydantic deprecations (mcp_use) + resource leaks; needs a warning-cleanup pass first
- **assert_called audit**: `test_mcp_client` upgraded to outcome assertions; the other ~160 files reviewed as legitimate spy assertions, bulk conversion deferred
- **Flake-gate script** ships but is not wired as a blocking gate (pre-existing flakes would red it) — wire after flake burn-down
- Pre-existing slow sleeps in `test_storage_bootstrap`/`test_lifecycle_core` noted for `slow`-marking

### Round 3 fixes (root-cause, per review)

- **Hermetic fence → explicit opt-in**: live-credential tiers now declare the keys they legitimately need via `HERMETIC_ALLOW_KEYS` in their own conftests (`tests/composio`, `tests/model_onboarding`); the fence honors declarations. Purpose documented in `tests/CLAUDE.md` (Environment & hermeticity).
- **Flake gate wired into CI**: a test that passes only on rerun fails the build (`run-tests-flake-gate.sh` wraps test-python).
- **Schemathesis isolated**: ephemeral free-port + `SCHEMA_FUZZ_PORT` override; mint retries; moved out of the parallel suite (its live server starves under shared-runner load) into `nx run api:test:schemathesis` + a serial CI step.
- **Chroma race guarded**: session pre-create of the canonical `EphemeralClient` makes the two bare call sites order-independent under pytest-randomly.

### Verification (round 3)

- **Full suite (CI-equivalent): 11,192 passed, 0 failed, 0 errors** (1:47)
- Collection audit: 465 test files / 11,567 tests — no orphans, no silent drops, no non-collecting files
- Schemathesis standalone: green; gaia-shared/tools-lints/vitest: green

### Follow-ups (round 4) — outcomes

- **Merged develop** (nurture email, local document parser, unsubscribe) — conflicts resolved (both alert-rules + semgrep lanes kept). Merged tree: **11,237 passed, 0 failed, 0 errors**.
- **RuntimeWarning-as-errors wired into CI** — exposed a real test bug (sync fire-and-forget scheduler mocked as AsyncMock); fixed at root (sync mock). Suite verified warning-clean with the flag.
- **`my_integrations` registry fallback: NOT a bug** — composio toolkits register categories by `tool_space` == integration id; the id-keyed lookup is correct. No change.
- **Slow sleeps**: the two deliberately-slow wedged-simulation tests marked `@pytest.mark.slow`.
- **assert_called audit**: scan verdict — call-asserts are predominantly paired with outcome asserts or are legitimate seam asserts; top offender already upgraded; no mass conversion (risk >> value).
- **Deferred with recommendation**: the 405-vs-422 undocumented-method shadowing is a real but low-severity API-surface wart; fixing means 405 handlers on ~40 ops — best done as its own API-hardening task, not in this test-infra PR. Same for documenting 429/403/307/410 in `responses=`.
- **Voice-agent + frontend (Playwright) tests**: explicitly deferred, per decision.

---

### Round 5 — the mutation-testing lane (diff-driven, CI-enforced)

**What it is.** Every changed `app/` module must ship tests that *notice* if the changed code is wrong. `mutmut` injects one tiny bug (a "mutant") at a time into a disposable copy of the module, runs the module's tests against each mutant, and classifies it: **killed** (a test failed — the suite would catch that bug), **survived** (all tests still pass — the suite runs but asserts nothing meaningful), or **no tests** (no covering test ran that line). The lane fails on survivors on lines the PR changed.

**Design decisions made along the way (each verified by a failing CI run before the fix):**
- **Self-contained script** (`scripts/ci/mutation-check.sh` + `scripts/test/mutation.sh` + `scripts/ci/mutation-matrix.py`): the `fromJSON`-matrix GitHub pattern proved broken in probe workflows (stale matrix + lane crash) — the lane now computes its own module→test-file matrix in-job, and runs modules in bounded parallelism.
- **AST-based test-file derivation**: patch targets in this codebase are strings (`patch("app.x.y.z")`), which grep misses; the scanner (`mutation-matrix.py`) collects references from imports + call-arg/assignment strings only — assert operands and fixture data no longer poison the scan. It skips `conftest.py` and non-`test_*.py` helpers, and follows one level of consumers (modules that import the changed module) so indirectly-tested modules still resolve.
- **Workdir copies, not symlinks**: a symlinked copy landed two paths to the same file on `sys.path` and CPython raised "cannot load module more than once per process" (observed in CI). Real copies fixed it.
- **`mutate_only_covered_lines` removed**: its coverage pass unloads every module imported during the stats run, and numpy/PyO3 C extensions cannot be re-imported in one process. The trampoline stats still scope each mutant to its covering tests, so runs stay cheap.
- **Diff-driven scoping**: mutmut has no "mutate only these lines" knob but honors `# pragma: no mutate` — the lane stamps the pragma onto every line outside the PR's changed ranges in the workdir copy. A 745-mutant module went from >28 min (CI runner, blew the job cap) to ~20 s. The survivor verdict remains as defense-in-depth.
- **Verdict is the source of truth**: mutmut can complete its work then hang at interpreter teardown (threads from C-heavy test runs) — a portable process-group timeout guards it, and interrupted runs fail via "not checked" mutants instead of silently passing.
- **Provably-equivalent mutants are filtered**: a mutant that only changes a `typing.cast()` type argument is a runtime no-op by contract — classified and noted, not failed. (The string-message class is *not* filtered — the lane caught a real message-pinning gap that way.)
- **Honest SKIPs, never silent passes**: mutmut 3.7 cannot mutate decorated functions (FastAPI endpoints, `@Cacheable`, ... — verified in its source), so modules whose changed lines are imports/decorators/statements skip with a printed reason; clean-run failures under mutmut's single-process phase transitions self-verify against the plain copy first.

**Real gaps the lane exposed — all fixed with tests that kill the mutants:**
- `_require_user_id` (memory endpoint) untested; error message pinned exactly
- postgresql: redundant ssl set removed (was an equivalent-mutant generator), exact URL/args/blank-value assertions, error-path + key-recording tests for `close_postgresql_db`
- google_meet registration return value pinned; `_redirect_to_oauth`, bearer/oauth routing, probe detection, OAuth discovery, `is_platform` default, `redirect_path` passthrough, wide-event log structure pinned in `test_integration_service.py`
- `noop_embedding.py` (new shared module) ships its own unit test — the lane's no-test-file rule rejected it on first push

**Lane outcome on this PR (CI, green):** 5 modules with mutatable changed lines → **all mutants killed** (runtime_adapter, noop_embedding, postgresql, integration_connection_service, mcp_client); 8 modules → SKIP with reason (changed lines are imports or inside decorated functions); 1 module (mongodb) → self-verified skip (module-level state doesn't survive mutmut's shared-process phases).

### Round 5 — every kind of test this branch runs, in plain words

Think of tests as nets, each catching a different kind of mistake. One net alone leaves gaps — a bug can fall between them. This branch layers the nets so that **every layer of failure has a net of its own**:

1. **Unit tests** — *one worker, one job, everyone else faked.* Test a single small piece of code in isolation. Catches: the logic *inside* one function breaking (wrong math, wrong branch, wrong return).
2. **Integration tests** — *two pieces talking, the outside world still faked.* Catches: the *handshake between layers* breaking — right idea, wrong shape, wrong format.
3. **Real-tier tests** — *the real database, real queue, no fakes at all.* Catches: the classic lie "works in the test but not with the real thing" — a fake that says yes where the real database says no.
4. **Contract tests** — *one app's promise to another app.* Catches: silently breaking another app's expectations (field renamed, response reshaped).
5. **E2E tests** — *the whole journey, from user message to final answer*, with a fake AI that writes down exactly what it was shown. Catches: the *plumbing between* the pieces breaking even though every piece is fine on its own.
6. **Prompt-contract tests** — *pin the exact message bytes the AI model receives.* Catches: a prompt change that looks right in the builder but gets mangled on the way to the model. (This PR caught a real one: the model was silently missing the memory/context frames because a middleware step threw them away — nobody had a net for it.)
7. **Stress tests** — *hammer the tricky concurrency spots* (background jobs, webhook replays, stream resumes, lock races). Catches: race conditions — the "works 99 times, breaks the 100th" bugs.
8. **Property tests** — *throw hundreds of random inputs at a function* and check it never breaks its own rules. Catches: the weird edge case no human would think to write a test for.
9. **Mutation tests** — *break the code on purpose and demand the tests notice.* Catches: tests that pass but prove nothing — the most dangerous kind. (Detailed above.)
10. **Regression-proof** — *your new test must FAIL against the old code.* Catches: a test that passes both before and after your change — i.e. it isn't testing your change at all.
11. **Diff-cover** — *every line you changed must be touched by a changed test.* Catches: changed code that no test ever runs.
12. **Flake gate** — *a test that only passes when rerun fails the build.* Catches: unreliable, order-dependent tests that poison trust in the whole suite.
13. **Meta tests** — *tests that watch the tests* (import fence, architecture invariants). Catches: the harness quietly breaking while everyone still says "green".
14. **Schemathesis** — *a robot that fuzzes the live API and checks it against its own docs.* Catches: the API drifting from what it promises (undocumented error codes, wrong statuses, shadowed routes — this PR documented all of it).
15. **Static gates** — ruff, mypy, semgrep, gitleaks, bandit. Catches: style, type and known-security-pattern mistakes before anything runs.

### Why this is objectively better (the short version)

Before this branch, the suite was mostly unit tests plus "it runs". A changed line could be touched by no test at all, a test could pass without checking anything, and a flaky test could lie about the suite's health. This branch closes each of those holes with a *mechanical* gate — no review discussion needed, the build itself refuses:

- **Changed code must be noticed when broken** → mutation lane (9)
- **Changed code must be covered by a changed test** → diff-cover (11)
- **A new test must actually bite** → regression-proof (10)
- **Tests must be trustworthy** → flake gate (12)
- **The live API must match its own docs** → Schemathesis (14)
- **The model must see the prompt you designed** → prompt-contract (6)

And the harness itself is watched by meta-tests (13), so the nets can't quietly rot.

### What it actually fixed (concrete finds, all with the failing test first)

- The model was silently receiving a prompt missing its memory/context frames (real production bug — found by the prompt-contract net)
- The device-pairing flow broke because of a deprecated library import that only the real-tier tests could see
- Every real-tier memory test failed on a chromadb contract change that unit fakes could never surface
- Mutation testing found whole functions with no meaningful coverage (`_require_user_id`, the OAuth redirect builder, the probe-detection branch) — each now pinned by a test that kills its mutants
- The flake gate caught a test that only passed by luck of ordering (an ambient trace stamp leaking into an exact-match assertion) — fixed at the root
- Schemathesis documented real API drift (undocumented 429 plan-gates, 405-vs-422 shadowing) — listed in the caveats below

### Round 5 — why some benchmark scores look poor (and what they actually mean)

**SonarCloud (60 issues, gate red on 1):** the PR's new code is *CI scripts and test files*, and SonarCloud's default ruleset measures them with production-code rules:
- 29× `shelldre:S7688` + 3× `S7677` + 2× `S131` — shell-style nits (`[` vs `[[`, `>&2`, missing `case` default) in the new CI scripts (`mutation.sh`, `regression-proof.sh`, `run-tests-flake-gate.sh`). These are style suggestions on operational bash, not defects; the scripts are gated by `bash -n` + the lanes they drive passing.
- 14× `python:S5778` (exception tests with multiple raising calls) + 7× `python:S9073` (composite assertions) — suggestions on *tests*: deliberately splitting every assertion would make the suite slower and noisier; the composite asserts that matter were already split where a failure needs pinpointing.
- 2× `python:S3776` (cognitive complexity) — the mutant-classifier in `mutation.sh` is a single focused script; splitting it would spread the logic.
- 1× `githubactions:S8541` — flags the lane's `uv sync` for omitting `--no-build`. **False positive**: the dependencies are the repo's own `uv.lock` (first-party, reviewed); `--no-build` genuinely cannot be used because `cloudinary==1.41.0` is source-only (no wheel) — verified by a red CI run. Marked for human confirmation in SonarCloud.
- The SonarCloud **security rating 3** is that single S8541; reliability/maintainability ratings are 1 (best). 59 of 60 issues are style/test-advice, not defects.

**Mutation lane's own "score" (why not every module shows 100% killed):** the lane can only mutate what `mutmut 3.7` can express. Decorated functions (every FastAPI endpoint) are skipped by the tool by design; the diff-driven scoping then leaves nothing mutatable when a PR's changed lines are imports or decorators — those modules SKIP with a printed reason, and their protection comes from the endpoint tests + diff-cover instead. Provably-equivalent mutants (cast-arg changes) are filtered with a note. Uncovered branches are reported as "no tests" notes. The number that matters is on the changed lines: **zero survivors**.

**Flake gate:** wired as blocking — it cost several red builds before the last flaky test (an ambient trace stamp leaking into an exact-kwargs assertion) was root-caused and fixed.


aryanranderiya:
## What this branch actually changes, and why the tests are worth having

**CI is green:** 30 checks pass, 6 skipped, 0 failures. `11242 passed, 2 xfailed` in 5:01 against live Postgres/Redis/Mongo/Chroma/RabbitMQ, plus 199 `gaia-shared` tests and the live-server contract gate. Coverage 85% against an 80% floor.

`164 files changed, +11,940 / −2,524`. Test churn: **64 added, 44 moved, 23 modified, 11 deleted.**

---

### The thesis

A test that cannot fail is theater, and "did you check it fails first?" is not enforceable by review — everyone says yes. The core of this branch is making that question **mechanical**.

### 1. Red-first, enforced by a machine (`scripts/ci/regression-proof.sh`)

The one I'd point at first. For every PR it creates a worktree at the base revision, **overlays only the PR's changed test files onto the old product code**, and runs them. If a changed test *passes* on base, it says so: the fix it claims to protect may not be needed, or the test never actually exercised the bug.

This is the FastAPI model, and it converts "I promise I saw it red" into a build step. It runs informational today (a gap-fill branch like this one legitimately adds tests for unchanged behaviour), with the enforcement condition written down rather than left implicit.

### 2. Flakiness is a defect, not a shrug (`scripts/ci/run-tests-flake-gate.sh`)

On failure the suite reruns **only the failures, once**. Pass-on-rerun **fails the build** (CPython's `--fail-rerun` lesson). A flaky test is normally the cheapest thing in the world to ignore; here it costs you the build.

This earned its keep immediately — see the pagination bug below, where the gate distinguished "genuine failure" from flake and was right.

### 3. Hermetic by default

A session fence blanks every credential-shaped env var, so no test can reach a live API or bill a real key, and the suite is deterministic regardless of a developer's `.env`. Tiers that genuinely need credentials declare them explicitly via `HERMETIC_ALLOW_KEYS` — nothing survives the fence by accident, only by declaration.

### 4. Canonical tiers

One place per purpose (`unit` → `integration` → `integration/real` → `contracts` → `e2e` → `stress` → `meta`), which is what the 44 renames are. `tests/CLAUDE.md` states the quality bar and the fixture catalog so the next person extends the right file instead of adding a rival one.

### 5. Ratchets that stop slow erosion

Suppression ratchet (diff-based per-file counts — no new `# noqa` / `# type: ignore` / `biome-ignore`), dead-code, docstring coverage, diff-coverage ≥90% on changed lines, and a live-server **schemathesis contract gate** that boots real `uvicorn` and fuzzes the OpenAPI schema over real HTTP.

---

### Why the new tests are good, concretely

- **`tests/stress/test_claim_race.py`** — 40 concurrent claimants race the tracked-todo execution lock. The docstring explains *why the fake is faithful* (its `set` has no `await` between check and write, mirroring Redis's single-threaded command atomicity), so the race resolves **deterministically**. No sleeps, no wall-clock. It states the invariant it protects: one runner in the critical section, losers never release a lock they didn't acquire.
- **`tests/unit/utils/test_property_*.py`** — Hypothesis sweeps all of `str` through parsers documented as having a defined outcome. Each module names **the fail-case the property guards** ("a parser change that lets a raw `httpx` error or a bare `AttributeError` escape would surface here").
- **`tests/meta/test_architecture.py`** — imports every tracked module under `app/` to catch circular-import and import-side-effect regressions in seconds.

The pattern: each says what would have to break for it to go red. That's the difference between a test and a coverage line.

---

### It has already paid for itself

Every one of these was found *by this infrastructure*, during this PR:

1. **`GET /api/v1/conversations?limit=55&page=191536681468777728` → 500.** The schemathesis gate found it on its first working run. `page` was bounded below (`ge=1`) but not above, and the services compute `skip = (page - 1) * page_size` — here `10534517480782774985`, past int64 max, so BSON can't encode the skip and the driver error escapes as a 500 instead of a 422. **Five endpoints shared the identical defect** (conversations, todos, memory, support, blog). The flake gate reran it and it failed identically, so this was correctly reported as genuine rather than dismissed.
2. **A cross-worker download race.** Every xdist worker shares one fastembed cache; on a cold first load they all pulled the same HuggingFace snapshot, and the loser handed onnxruntime a `model.onnx` that hadn't finished downloading. 24 memory tests errored intermittently. Now serialized by a file lock, mirroring the advisory lock `pg_engine` already takes around `create_all`.
3. **An untested startup path.** `_ensure_timestamptz_columns` runs on every boot and had **no test at all**. Writing one exposed that it interpolated identifiers unquoted — fine for today's all-lowercase whitelist, fatal the moment an entry needs quoting. Now uses `DDL` with the dialect's identifier preparer, and clears a semgrep finding that had been sitting on `develop`.
4. **13 lint suppressions, two of which hid real defects** — an async-generator fixture annotated `-> CheckpointerManager` instead of `AsyncIterator[...]`, and a `noqa: BLE001` that was discarding the traceback in the import smoke test (the frames are the entire point for a circular-import chain). The rest were simply wrong.
5. **The dead-code gate itself was broken** — it gated on knip's *output being non-empty* with stderr merged in, so an unrelated Node deprecation warning failed the lane while printing "0 dead code items found", a contradiction no code change could clear.

---

### Known limits, stated rather than buried

- **`regression-proof` is informational**, not blocking — gap-fill tests legitimately pass on base. The enforcement condition is documented in the job.
- **The semgrep lane is informational.** `apps/api/app` is now at **0 findings across 815 files** and the scanner image is digest-pinned, but `--config p/python` still fetches rules from the registry per run, so enforcing it would let an upstream rule change red unrelated PRs. Pinning the ruleset is the prerequisite.
- **The schemathesis gate's green scope is deliberately narrow** (`SCOPED_OPERATIONS`). The module docstring lists the contract drift found by the full-schema exploratory pass — undocumented 429 plan-gates, 405-vs-422 route shadowing, undocumented 403/307/410 — **reported, not silenced**, and tracked for the API-hardening follow-up.
- The gate needs the app's own startup preconditions (`ai_models` / `subscription_plans` non-empty, enforced by `startup_validation`). CI's Mongo is empty, so the fixture seeds the minimum each check counts — only into an empty collection, removing only what it inserted, so a developer's real catalog is untouched.

aryanranderiya:
## Follow-up: the last two informational lanes are now enforced — and one of them was fake

**All 19 quality-gate lanes are enforced.** CI: 30 pass, 6 skipped, 0 failures.

### regression-proof was not informational, it was vacuous

It has been reporting success **without ever running a test**. The script redirected pytest into `/tmp/opencode/` — a hardcoded path from an unrelated tool that does not exist on a GitHub runner:

```
regression-proof.sh: line 67: /tmp/opencode/regression-proof.log: No such file or directory
regression-proof: all changed tests fail on base (rc=1) as required   ← "success"
```

Redirect fails → pytest never runs → `rc=1` skips both failure branches → falls through to printing success. Worse than no gate, because it looked like coverage.

Fixing that exposed two more bugs stacked behind it, each of which would independently have kept it fake:

1. `--strict-markers` ran against the **base** `pytest.ini`, which rejects markers this branch adds (`regression`, `stress`) — every run aborted during collection.
2. pytest ran from the worktree root, but `app_factory` mounts `StaticFiles(directory="app/static")` on a **CWD-relative** path — so every test that builds the FastAPI app *errored* rather than failing. An error counts as "did not pass", so the lane would have kept claiming proof it never obtained.

**Why it can be enforced now.** The old rule — *every* changed test must fail on base — is only true of bug-fix PRs; gap-fill branches legitimately add tests for behavior base already gets right, which is exactly why it was parked. It is now scoped to `@pytest.mark.regression`, opt-in **per test, not per file**: of the four timestamptz tests, three pin the quoting bug and one covers the pre-existing skip path, and only the three are marked. A diff with no such tests passes trivially.

Verified **both directions**: the marked tests all fail on base, and marking a test that *passes* on base makes the gate exit 1.

Here is the lane doing its job in CI, naming the exact production failure each test pins:

```
FAILED test_promotes_naive_column_whose_name_requires_quoting
       → column "probecreatedat" does not exist
FAILED test_list_rejects_page_that_would_overflow_the_mongo_skip
       → assert 500 == 422
4 failed, 14 deselected, 3 errors
```

It also now prints per-test outcomes and distinguishes `FAILED` (a real assertion) from `ERROR` (weaker evidence — the test may not have run at all), instead of a bare count.

### semgrep: rules vendored, then enforced

Findings were already burned down, but pinning the image fixed the *engine*, not the *rules* — `--config p/python` re-resolved the ruleset from the registry every run, so what CI enforces could change with no commit behind it. `tools/semgrep/p-python.yaml` now pins the pack (151 rules, matching the count CI reported against the registry); `refresh-p-python.sh` reproduces the fetch so updates arrive as reviewable PRs.

Equivalence was verified, not assumed: same 815 targets, same 151 rules, 0 findings — **and** it still flags the pre-fix `_ensure_timestamptz_columns` pattern, proving it detects rather than silently matching nothing.

Kept as a whole-directory scan rather than diff-scoped like ruff/mypy: a security lane that only reads files a PR happens to touch stops noticing everything else.

### The pattern worth naming: these gates fail open

Three separate lanes in this branch reported success while doing nothing — regression-proof's `/tmp/opencode`, the dead-code lane counting stderr chatter as findings, and (nearly) semgrep. The last one almost shipped: locally the pinned image reports `Targets scanned: 0` **and exits 0**, because this is a git *worktree* whose `.git` is a file outside the Docker mount, so git-tracked discovery finds nothing. CI does a real checkout and scans 815.

None of these were caught by their exit codes, because a broken gate's exit code is `0`. They were caught by asking *"did this actually run?"* — so regression-proof now fails when pytest produces no summary line, and target counts are checked rather than trusted.

### Test accounting, for the record

| | Count |
|---|---|
| Tests that exist | 11,617 |
| Selected by CI's marker filter | 11,244 |
| Actually ran | 11,244 (11,242 passed + 2 xfailed) |
| Skipped at runtime | 0 |

The 373 deselected reconcile exactly: 371 `composio` (live credentials), 1 `model_onboarding` (bills real tokens), 1 `schemathesis` (own serial step, passes there). With the separate steps, CI executes **11,444** tests per run.

PR #972, opened 2026-08-10
fix(agent): stop checkpointed threads accumulating a prompt stack per run

aryanranderiya:
## Problem

Prod's `langgraph` DB hit **19 GB** (checkpoint_blobs 11 GB, checkpoint_writes 7.9 GB) and is growing **~2.4 GB/day since Aug 6** — despite #818's DeltaChannel fix (deployed, langgraph 1.2.7). Evidence from prod:

- The recent giant writes are **~4.8 MB SystemMessage batches** containing **39 copies of the comms system prompt** on one recurring workflow thread.
- Every run injects a fresh ~125 KB prompt stack; `manage_system_prompts_node` filters stale copies **from the per-call view only** (its own docstring: "the persistent checkpoint state still grows unfiltered").
- The `end_graph_hooks` node returned the **full hook state** as its update, re-serializing the entire accumulated message list into `checkpoint_writes` every run.
- `__pregel_tasks` (5 GB) mirrors the same payloads in Send packets.

## Fix

1. **Durable pruning**: `manage_system_prompts_node` relays the ids of slot-stale messages via an in-memory key (`PRUNED_MESSAGE_IDS_KEY`, popped before any channel write); the model node emits `RemoveMessage` tombstones so each prompt slot holds exactly one message *in the checkpoint*, not just per-call.
2. **No more full-state echo**: the end-hooks node returns only the keys its hooks actually changed (identity diff). Its hooks (follow-up streaming, memory ingestion) are side-effecting and never change state, so this is normally an empty update.

## Verification (failing-first, both tiers)

- `tests/e2e/test_checkpoint_state_growth.py` — real comms graph: stale prompt copies are tombstoned out of the thread (was 3 copies after 3 runs → 1), conversation turns survive, end-hooks node emits no messages write. Red before the fix, green after.
- `tests/integration/real/test_postgres_checkpointer.py::TestCheckpointStorageStaysBounded` — **real Postgres checkpointer + real comms graph**, measures actual stored bytes for the thread across 4 recurring runs: red on master (super-linear, ~3.5× over a linear budget and growing per run), green here (~1 stack per run). Full e2e tier: 507 passed.
- Not exercised: real prod LLM/tools (fake scripted model), and the fix does not shrink existing data — see cleanup note below.

## Deliberately out of scope (needs own design)

- Each run's graph INPUT still re-sends the full Mongo-sourced history into the thread (double bookkeeping between Mongo and the checkpointer).
- No checkpoint retention/TTL exists — old threads live forever. Existing 19 GB needs a one-time cleanup + VACUUM regardless of this fix.
- `deep_research` previously produced up to 71 MB single ToolMessages (pre-compaction era); compaction middleware now bounds tool outputs, but `__pregel_tasks` payloads deserve a look.

https://claude.ai/code/session_012FxQb4So3EUubqhrWXoP7r

### DIFF
--- a/apps/api/app/override/langgraph_bigtool/create_agent.py
+++ b/apps/api/app/override/langgraph_bigtool/create_agent.py
@@ -33,7 +33,13 @@ from typing import Any, cast
 
 from langchain.agents.middleware import AgentMiddleware
 from langchain_core.language_models import LanguageModelLike
-from langchain_core.messages import AIMessage, HumanMessage, ToolCall, ToolMessage
+from langchain_core.messages import (
+    AIMessage,
+    HumanMessage,
+    RemoveMessage,
+    ToolCall,
+    ToolMessage,
+)
 from langchain_core.runnables import Runnable, RunnableConfig
 from langchain_core.tools import BaseTool, StructuredTool
 
@@ -70,6 +76,7 @@ from app.override.langgraph_bigtool.hooks import (
     sync_execute_hooks,
 )
 from app.override.langgraph_bigtool.utils import (
+    PRUNED_MESSAGE_IDS_KEY,
     RetrieveToolsResult,
     State,
     dedupe_str_list,
@@ -96,6 +103,45 @@ def _prepare_fallback(
     if fallback_llm is None or is_default_model_config(model_configurations):
         return None
     return lambda: fallback_llm.bind_tools(tools_to_bind)  # type: ignore[attr-defined]
+
+
+def _pop_pruned_tombstones(state: State) -> list[RemoveMessage]:
+    """Tombstones for the slot-stale messages the pre-model hooks dropped.
+
+    ``manage_system_prompts_node`` filters old prompt copies out of the
+    per-call view and relays their ids under ``PRUNED_MESSAGE_IDS_KEY``; the
+    model node emits these tombstones in its own channel update so the pruning
+    also reaches the checkpoint. Popped, never forwarded — the key is a relay,
+    not a state channel.
+    """
+    pruned = cast("dict[str, Any]", state).pop(PRUNED_MESSAGE_IDS_KEY, None) or []
+    return [RemoveMessage(id=message_id) for message_id in pruned]
+
+
+_UNSET = object()
+
+
+def _changed_hook_keys(before: State, after: State) -> State:
+    """Only the keys the end-graph hooks actually changed, by identity.
+
+    The end hooks are side-effecting (follow-up streaming, fire-and-forget
+    memory ingestion) and pass state through; echoing the full state back as
+    the node's update re-serializes the entire ``messages`` list into
+    ``checkpoint_writes`` on every run — on a long-lived workflow thread that
+    was ~4.8 MB per run of pure duplication. Identity (not equality) so a hook
+    that rebuilds the dict without touching the values still counts as
+    unchanged, and a genuinely new value always survives.
+    """
+    if after is before:
+        return cast("State", {})
+    return cast(
+        "State",
+        {
+            key: value
+            for key, value in after.items()
+            if cast("dict[str, Any]", before).get(key, _UNSET) is not value
+        },
+    )
 
 
 def create_agent(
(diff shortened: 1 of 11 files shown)


apps/api/app/override/langgraph_bigtool/create_agent.py
```diff
@@ -98,6 +105,45 @@ def _prepare_fallback(
...
+    not a state channel.
+    """
+    pruned = cast("dict[str, Any]", state).pop(PRUNED_MESSAGE_IDS_KEY, None) or []
+    return [RemoveMessage(id=message_id) for message_id in pruned]
+
+
+_UNSET = object()
+
+
+def _changed_hook_keys(before: State, after: State) -> State:
```
aryanranderiya: this function shouldnt be here

apps/api/app/override/langgraph_bigtool/create_agent.py
```diff
@@ -98,6 +105,45 @@ def _prepare_fallback(
     return lambda: fallback_llm.bind_tools(tools_to_bind)  # type: ignore[attr-defined]
 
 
+def _pop_pruned_tombstones(state: State) -> list[RemoveMessage]:
```
aryanranderiya: this function too

apps/api/app/override/langgraph_bigtool/create_agent.py
```diff
@@ -98,6 +105,45 @@ def _prepare_fallback(
...
+    per-call view and relays their ids under ``PRUNED_MESSAGE_IDS_KEY``; the
+    model node emits these tombstones in its own channel update so the pruning
+    also reaches the checkpoint. Popped, never forwarded — the key is a relay,
+    not a state channel.
+    """
+    pruned = cast("dict[str, Any]", state).pop(PRUNED_MESSAGE_IDS_KEY, None) or []
+    return [RemoveMessage(id=message_id) for message_id in pruned]
+
+
+_UNSET = object()
```
aryanranderiya: what does this do? is it the right place and best method? global vars are usually a bad idea

apps/api/app/override/langgraph_bigtool/utils.py
```diff
@@ -110,6 +110,16 @@ class State(_BigtoolState):
...
 
 
+# In-memory relay from `manage_system_prompts_node` to the model node: ids of
+# slot-stale messages (old system prompts / clock lines) the hook dropped from
+# the per-call view. The model node turns them into RemoveMessage tombstones in
+# its own channel update so the pruning also reaches the CHECKPOINT — without
+# this, every run's prompt stack stays in the thread forever (a production
+# workflow thread accumulated 39 copies, ~4.8 MB of checkpoint writes per run).
+# Never a state channel: the model node pops it before returning its update.
+PRUNED_MESSAGE_IDS_KEY = "_pruned_message_ids"
```
aryanranderiya, line 116: wont this keep accumulating?

apps/api/app/override/langgraph_bigtool/create_agent.py
aryanranderiya: Moved — `pop_pruned_tombstones` now lives in `utils.py` beside `PRUNED_MESSAGE_IDS_KEY` (the constant it serves), imported into `create_agent.py`.

apps/api/app/override/langgraph_bigtool/create_agent.py
aryanranderiya: Moved — `changed_hook_keys` now lives in `hooks.py` with the other hook utilities.

apps/api/app/override/langgraph_bigtool/create_agent.py
aryanranderiya: `_UNSET` was a missing-vs-None sentinel for the dict diff. Agreed it wasn't needed — removed entirely; the diff now uses `key not in before` which needs no sentinel.

apps/api/app/override/langgraph_bigtool/utils.py
aryanranderiya: No — it can't accumulate. The hook OVERWRITES the key with this call's stale ids (never appends), and the model node pops it in the same superstep. After the tombstones apply there's nothing stale left, so subsequent calls relay an empty list. It's never persisted: popped before the node returns its update, and if a malformed value ever appears it raises instead of silently passing through.

PR #976, opened 2026-08-10
chore(mypy): maximum strictness — no explicit Any anywhere in the API

aryanranderiya:
## Summary

Every strict mypy flag is now on for the whole API, including `disallow_any_explicit` — so `Any` can no longer be written anywhere in `apps/api/app` without an explicit, reviewable exception. Roughly 450 files were retyped to get there: `Any` became `object` or a precise type, and the places that then needed narrowing got a real `isinstance` check instead of a silent assumption. The strict-mypy "island" that previously covered a handful of modules is gone, because the flags it applied are now global.

## Why

The island was the problem. Strict checking applied to about twenty hand-listed modules, so the rest of the API could pass `Any` around freely — and `Any` launders a wrong type through every function it touches, turning what should be a type error into a runtime one somewhere far away. A codebase where the guarantee holds in some directories and not others gives you the cost of strictness without the benefit: you still cannot trust a signature you did not personally verify.

### Type checking

- Global: `disallow_any_explicit`, `disallow_any_generics`, `disallow_untyped_calls`, `disallow_untyped_decorators`, `disallow_subclassing_any`, `strict_optional`, `strict_equality`, `extra_checks`, `warn_redundant_casts`. The per-module override block is deleted as redundant.
- Test modules are included in `disallow_untyped_calls` — an untyped call from a test is the same hole as one from app code.

### Code

- `Any` → `object` or a precise type across ~450 files, with `isinstance` narrowing (and `cast` to a `TypedDict` after narrowing) where the value is then used concretely.
- `settings.py` hoists roughly forty env-var fields that were duplicated across the production and development subclasses up into `CommonSettings`, so `get_settings()` can return a real type instead of `Any`. Same defaults, same inheritance — one definition instead of three, which is also why a field added to only two of the three could previously go unnoticed.

### About the 837 new ignores

818 of them are `explicit-any` on a single line: the `class Foo(BaseModel)` declaration itself. Pydantic's and SQLAlchemy's base classes are generic in `Any`, so subclassing them trips the flag no matter what the subclass looks like. These are annotations of a framework fact, not suppressed findings — the alternative is not turning the flag on. The remaining ~19 are third-party boundary seams (SQLAlchemy declarative base, langgraph_bigtool, livekit, async Redis) and each is a specific error code, never a bare ignore.

The suppression baseline roughly doubles as a direct consequence. That number is the honest cost of the flag and it is worth stating plainly rather than burying: the ratchet is supposed to shrink over time, and this is a one-time step up that only shrinks again if pydantic's typing improves.

---

## The bug

Retyping a codebase this size will change behaviour if you let it, and it did in nine places. A review pass found them and each is fixed here, with a test seen failing first. They share one shape: a type narrowing was added, and the branch where the narrowing *failed* silently did nothing instead of failing loudly.

**Symptom** — the worst one was user-visible and wrong rather than broken: `CUSTOM_ADD_RECURRENCE` guarded the fetched calendar event with an `isinstance` check, and when it was not a dict, skipped attaching the recurrence rule, sent the unmodified event anyway, and returned success naming the rule the user asked for. The agent told the user their event now repeats weekly. It does not.

**Root cause** — narrowing added to satisfy the checker, with the else-branch left empty. Master raised a `TypeError` here (`calendar_tool.py`, `CUSTOM_ADD_RECURRENCE`).

**The fix** — raise. The other eight are the same correction:

- `hil/utils.py::current_tool_calls` stopped at the first `AIMessage` instead of walking back to the last one that *has* calls — and every `AIMessage` carries `tool_calls=[]`. It returned no siblings, which skipped the double-run guard in `policy.py`. That guard protects an approval gate.
- `linear_tool` turned a null estimate into `0.0`, reporting an unestimated issue as estimated at zero points — the same bug commit `3765daa40` already fixed once for `priority`, left on the sibling fields.
- The artifact registry stored `mtime` as `0.0` but still compared it against the raw `None`, so an artifact with no mtime was re-published and re-delivered on every event instead of being skipped.
- Four onboarding read sites gained `model_validate` over legacy Mongo rows — in the same PR that adds `SkipValidation` to that field precisely because those rows cannot validate. A pre-constraint profession aborted the whole pipeline: no todos, no writing style, no first message.
- `dict_bag` was used as a write target in three places, against its own docstring ("Never use these to write"): it returns a throwaway `{}` when the key is absent, so a weather country, a trigger config field and a numbered-list ordinal were written to something discarded.
- `chat/state.py` had the same shape — an appended todo-progress card written to a local list that was never stored back.
- `tool_data_utils` began popping the key for an explicit empty `tool_data: []`, where master left `[]` in place, so consumers indexing it get a `KeyError`.
- `bot.py` returned `""` for a non-datetime `created_at` instead of raising, showing a blank join date. It now raises an `AppError` naming the field and the type it got, rather than a bare exception — the observability gate requires a structured error with a `why` and a `fix`, and a handler that fails loud without explaining itself is only half the job.

**Why the suite missed it** — these paths were covered, but only on their happy branch. No test passed a non-dict event, an `AIMessage` with no calls, a null estimate, or a legacy onboarding row. The tests now do, and none of the fixes needed a type hole: mypy stays at zero errors.

## How to verify

1. `cd apps/api && uv run mypy app` — zero errors, and confirm it now covers every module rather than the old island (delete a real annotation somewhere outside the former island and check it goes red).
2. `uv run pytest tests/unit` — 10,832 passing.
3. For the behaviour fixes: run the calendar recurrence tool against a proxy that returns a non-object payload and confirm it raises rather than reporting success; construct an approval state whose trailing `AIMessage` has no tool calls and confirm the sibling guard still sees the real call set.

**Not verified:** the retyping is mechanical across ~450 files and only the paths with tests are proven. The nine fixes above are the ones a review found; a diff this size cannot be exhaustively reviewed by reading, so the mutation lane and the unit suite are doing most of the work.

## Risk

This is the highest-risk PR of the open set, and the reason is stated above: the danger is not the type checker, it is the edits made to satisfy it. Nine such edits were found and fixed; the honest position is that the review pass covered the surfaces most likely to hide them rather than every one of 517 files. The failure mode to watch after merge is a value that used to raise now quietly becoming `0`, `""` or `{}` — so the first thing to check on any anomaly is whether the code path recently gained an `isinstance` guard.

`settings.py` is the other one worth a second look: it is the only structural change, and a field that failed to hoist correctly would read as unset rather than erroring.


aryanranderiya:
@greptileai please compare every single file, comprehensively review each file before and after. i don't want type checking to bite us in the ass. find out any issues that have introduced by adding strict typing. any logical or functional issues too.

aryanranderiya:
@coderabbitai please compare every single file, comprehensively review each file before and after. i don't want type checking to bite us in the ass. find out any issues that have introduced by adding strict typing. any logical or functional issues too. i dont want minor issues at all. i want actually breaking issues

aryanranderiya:
@coderabbitai

Branch is now merged with `master` and re-pushed. Before you start: your last two attempts were skipped outright ("486 files exceed the limit of 100"), so a plain full-PR review will not run. Please review **incrementally instead of the full diff** — `@coderabbitai review` against the latest commits only, and if that still exceeds the limit, walk the commits one at a time.

What I actually want checked — not style, not nits. This PR converts ~1,100 sites from `Any`/`dict[str, Any]` to `dict[str, object]` plus checked accessors in `apps/api/app/utils/json_helpers.py` (`text_bag`/`int_bag`/`float_bag`/`bool_bag`/`dict_bag`/`list_bag`). Every one of those accessors **silently returns a default when the runtime type does not match the `isinstance` check**. That is the failure mode: no exception, no log, just a wrong value.

One instance of exactly this has already been found and fixed in `3765daa4`:

> `int_bag(issue, "priority")` on Linear issues. Linear's published schema types `Issue.priority` as `Float!`, so `2.0` is a valid wire value. `isinstance(2.0, int)` is `False`, so `int_bag` returned `0` and every issue rendered as "no priority". The pre-PR code (`issue.get("priority", 0)` into a dict lookup) was correct because `2.0` and `2` hash equal. Fixed by reading with `float_bag`.

**Please hunt for more of that same class**, prioritising these:

1. `int_bag` on any field an external API may serialise as a float (JSON numbers, GraphQL `Float`, anything divided or averaged upstream). Call sites: `twitter_tool`, `linkedin_tool`, `google_docs_utils`, `weather_utils`, `notion_md`, `urgency_tool`, `research_utils`.
2. `bool_bag` on any field an API sends as `0`/`1` or `"true"`/`"false"` — `isinstance(1, bool)` is `False`, so it returns the default.
3. `text_bag` on any field that can be a number, `ObjectId`, `datetime`, or `None` where the pre-PR code did `str(...)` — the old code stringified it, the new code returns `""`.
4. `dict_bag`/`list_bag` where the payload can legitimately be the other container, or where the old code let `None` through and a downstream `if x is None` branch now sees `{}`/`[]` instead.

For each finding, please state the concrete wire value that triggers it and what the pre-PR code did with that same value. A finding without both of those is not actionable. If a call site is provably safe because we own the producer, skip it — I only want the ones where the payload is foreign.

aryanranderiya:
@coderabbitai review

Re-requesting on the current head — the branch has moved a lot since your last pass (master merged twice, 11 further commits). As before: **review incrementally, not the full diff** — a whole-PR review is skipped at 487 files, so walk the recent commits instead.

Since your last look, these landed and are the highest-value things to check:

- `3765daa4` — **a real bug**: Linear's schema types `Issue.priority` as `Float!`, so `2.0` is a valid wire value; the migration read it with `int_bag`, whose `isinstance(value, int)` check rejects a float and fell back to `0`, rendering **every** issue as "no priority". Fixed by reading with `float_bag`.
- `86394267b` (author's) — restores several behaviours the strictness pass had quietly turned from loud failures into silent ones: `dict_bag` used as a *write* target in three places (the write vanished into a throwaway `{}`), a null Linear estimate becoming `0.0`, artifacts comparing stored `0`/`0.0` against a raw `None`, and `tool_data: []` being popped so consumers indexing it got a KeyError.
- `a0627df72`, `67abc8c40` — observability/structured-error fix plus three `@pytest.mark.regression` markers that were not actually red on base.
- `c294642cd` — first tests for `process_webhook`'s envelope reads.

**What I most want a second pair of eyes on:** the accessor family in `apps/api/app/utils/json_helpers.py` silently substitutes a default whenever the runtime type does not match its `isinstance` check. The two bug classes found so far are both instances of that:

1. **wrong numeric type** — `int_bag` on a field the provider sends as a float (the Linear priority bug), and
2. **nullable collapsed** — a non-optional accessor on a field that is legitimately null, so "unknown" becomes `""`/`0`/`0.0`/`False`.

I swept class 2 across all 36 sites where a `.get("k")` with no default became a non-optional accessor and found nothing further — Linear's remaining reads are of non-null fields (`Cycle.number: Float!`, `Issue.priority: Float!`, `IssuePayload.success: Boolean!`), and the rest are truthiness-guarded on the next line. **Please check that conclusion**, and hunt class 1 (and `bool_bag` on `0`/`1` fields, where `isinstance(1, bool)` is `False`) across the non-Linear integrations: gmail, notion, twitter, linkedin, google-docs, weather, todoist.

For each finding, state the concrete wire value that triggers it and what the pre-PR code did with that same value. Skip anything where we own the producer — I only want foreign payloads.

PR #1054, opened 2026-08-18
refactor(chat): one splitter and one token constant for message breaks

aryanranderiya:
## Summary

Splitting an assistant reply into chat bubbles had two implementations and three hand-written copies of the token that triggers it. This collapses both to one. No behaviour change — it removes the ways a future change can silently pick the wrong one.

## Why

`splitMessageByBreaks` split on `<NEW_MESSAGE_BREAK>` but knew nothing about OpenUI fences. `splitByBreaksPreservingFences` did the same job and additionally refused to bisect a `:::openui` block. Both were exported from the same barrel, so every caller had to know which one was correct — and the web bubbles decided inline, with the same `text.includes(":::openui")` check duplicated in two components, while mobile just called the fence-aware one unconditionally.

Separately, the token string was typed out by hand in `openui-parser.ts` and in `first_message_service.py`, in spite of a constant existing for exactly this whose own doc comment asks callers to use it. Nothing keeps a hand-written copy in step with the real one, and the drift would surface only as a reply that quietly stops splitting into bubbles.

### Web / Mobile / Shared

- One splitter. The fence-aware implementation is a strict superset — given text with no fence it performs the identical split — so it is now the only one, renamed to `splitMessageByBreaks`. The `:::openui` branch disappears from `ChatBubbleBot` and `TextBubble`.
- `openui-parser.ts` imports `NEW_MESSAGE_BREAK_TOKEN` instead of spelling the literal.
- The web-side `messageBreakUtils` no longer re-exports the splitter — that gave one function two import paths and nothing else. The file now owns only the Framer Motion timings it actually defines.

### API

- `first_message_service.py` imports `NEW_MESSAGE_BREAKER` instead of hardcoding the token in the fallback greeting.

## How to verify

1. `pnpm nx run-many -t type-check --projects=web,mobile` and `pnpm nx run bots-e2e:test`.
2. Open a chat and send something that produces a multi-bubble reply — it should still render as separate bubbles with the staggered reveal, on web and mobile.
3. Send a prompt that produces an OpenUI component alongside text. The `:::openui` block must stay in one bubble rather than being cut in half — this is the case that previously depended on picking the right splitter.

**Not verified:** the onboarding fallback greeting in `first_message_service.py` only fires when first-message generation fails, which was not triggered; the change there is a literal-for-constant substitution producing a byte-identical string.

## Risk

Contained to how already-generated text is divided for display. The consolidated splitter is the one mobile and the OpenUI path already used, and web text with no `:::openui` in it takes the same branch it did before, so the realistic failure mode would be a bubble boundary landing differently in text that contains a fence — visible immediately in the chat UI, and revertible on its own since nothing else depends on this commit.

### DIFF
--- a/apps/api/app/services/onboarding/first_message_service.py
+++ b/apps/api/app/services/onboarding/first_message_service.py
@@ -10,6 +10,7 @@ from app.agents.prompts.onboarding_prompts import (
     FIRST_MESSAGE_GENERATION_PROMPT_GMAIL,
     FIRST_MESSAGE_GENERATION_PROMPT_NO_GMAIL,
 )
+from app.constants.general import NEW_MESSAGE_BREAKER
 from app.constants.log_tags import LogTag
 from app.models.onboarding_models import (
     ClarifyAnswerRecord,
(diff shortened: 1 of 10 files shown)


aryanranderiya:
Folding this into #1051 — one PR for the whole message-splitting change instead of two.

PR #1057, opened 2026-08-19
refactor(agents): frame internal agent handoffs in XML tags

Dhruv-Maradiya:
## Summary

GAIA's tiers hand work to each other as plain text that lands in the next tier's context: the background executor's result reaches the user-facing comms agent, subagent results reach the executor. That text used to be prefixed with a bare marker like `[EXECUTOR_RESULT]`, which marks only where an internal payload starts and leaves the model to infer where it ends. Every cross-tier payload is now wrapped in an XML-style block with an explicit open and close tag, and a single enum owns both the vocabulary and the framing, so the code that writes a tag and the code that strips one cannot drift apart.

## Why

A prefix gives the model no end boundary, and a bracketed word reads like ordinary text worth repeating — comms occasionally copied one into a user-facing reply. Subagent results had a sharper problem: they were joined with a literal `---`, which a subagent's own report can contain, so the executor could not reliably tell one report from two. The delivery instructions prepended to a result carried no marker at all and ran straight into the payload.

### API

- `AgentTag` and `wrap_agent_payload` (`app/constants/agents.py`) own the tag vocabulary *and* how a payload is framed. `strip_internal_agent_tags` builds its pattern from the enum, so adding a tag extends the backstop for free instead of requiring a second edit someone will forget.
- Executor → comms handoffs (result, error, cancellation), the card-suppression note, and the bubble-split delivery note are now framed blocks.
- Subagent → executor results are `<subagent_result agent="…">` blocks, replacing `[gmail result]` joined by `---`.
- The comms prompt, the prompt-contract clauses, and the eval gate that derives its inputs from the prompt moved with the vocabulary. The gate now also catches a lone closing tag, which the opening-only check scored clean.
- Duplicate marker constants in `app/constants/executor.py` deleted; nothing imported them.
- The cancellation record's test now pins the whole record rather than one phrase of it, so the denial text that stops comms claiming a cancelled task finished can no longer be rewritten unnoticed.

## Where the framing is and is not scrubbed

Comms re-voices a framed payload on the background narration path, and that path strips the framing from its own output before delivery — the payload is in the turn being re-voiced, so an echo there is always a mistake.

The interactive chat turn deliberately does **not** strip. An unconditional strip also deletes a tag the user legitimately asked about ("what does `<executor_result>` mean") — intact on the live stream, silently missing on reload. Distinguishing an echoed tag from a deliberate one is not possible from the text alone, and measurement says the trade is not worth making: driving the real comms graph with the production model, the framing was never spontaneously reproduced across 54 delivery turns. It only appears when a user explicitly asks for the raw context, which is exactly the case where deleting it is wrong.

## How to verify

1. `mise dev --sim`, then `mise seed`.
2. `POST /api/v1/chat-stream` with message `add a todo [[tool:create_todo {"title":"framing check"}]] [[say:Created the todo.]]`.
3. Read the comms thread's checkpoint blob out of Postgres (`checkpoint_blobs`, `thread_id` = the conversation id). The message comms received is `<delivery_instructions>…</delivery_instructions>` followed by `<executor_result>Created the todo.</executor_result>` — one framed block per concern, each closed.
4. In Mongo, the saved turn carries comms' re-voiced reply with no framing in it.
5. Subagent framing under real infra: `nx run api:test:real` covers the join's `<subagent_result agent="…">` blocks through real Postgres/Redis/Mongo (`tests/e2e/test_hil_barrier_e2e.py`).

**Not verified:** behaviour beyond the tag rule — tone, delegation — was not put through the eval suites. The real-model run that exercised a live executor result hit the free-plan usage cap, so the payload it re-voiced was a one-liner rather than a large structured result.

## Metrics

The premise that models copy XML tags less than bracket markers did **not** hold up when measured, so this change rests on the explicit payload boundary and the single shared vocabulary, not on a model-side win.

| Metric | Bracket markers | XML tags | How measured |
| --- | --- | --- | --- |
| Framing echoed on a normal delivery turn | 0/54 | 0/54 | Real comms graph + real production model, graded on raw output before any strip |
| Framing echoed when the user demands the raw context verbatim | 7/12 | 7/12 | Same harness, follow-up turn asking comms to "copy it exactly including any tags" |

Method: the "before" arm ran from a worktree at `45e843d4b` under this branch's environment, so it had the old markers *and* the old prompt; both arms were graded by one pattern covering both vocabularies. The probe is a throwaway script and is **not committed**, so these figures are reported, not reproducible from this diff. The second row is comms complying with a direct request for its context, which the framing style does not affect.

## Risk

The comms prompt changed wording in six places, so it touches every chat turn; the tag rule itself is gated deterministically, but nothing here measures the rest of the prompt's behaviour. The subagent join's output format changed, so anything parsing the old `[agent result]` shape would break — the executor consumes it as prose and nothing else reads it.


Dhruv-Maradiya:
✅ **Confirmed complete and merged.**

This is the "wrap context blocks in HTML-like tags" work (wrap-comms-context). It was merged to `master` on **2026-08-19** by @Dhruv-Maradiya via merge commit `14b6906dd999945eb4d5f751c4b0e3243ea1cfb9`. Executor → comms handoffs and subagent results are now framed in XML-style open/close tags (`<executor_result>…</executor_result>`, `<subagent_result agent="…">…</subagent_result>`, `<delivery_instructions>…</delivery_instructions>`), with a single `AgentTag` enum owning the vocabulary so comms stops surfacing those blocks to users.

Marking as done.

PR #1061, opened 2026-08-20
chore(lint): re-enable A001/A002/B904/S608, add complexity ratchet

Dhruv-Maradiya:
## Summary

Re-enables four ruff rules that were globally switched off and fixes every violation they found: builtin-name shadowing (`id`, `type`, `filter`, and friends), exceptions re-raised without chaining to their original cause, and a SQL-injection heuristic that had two stale false positives. Also enables complexity rules (too many returns/branches/arguments/statements), but since fixing all of today's ~240 pre-existing violations would mean refactoring 180 files, a new lint (`tools/lints/check_plr_complexity.py`) grandfathers today's debt per file — quietly, until a PR touches that file, at which point its complexity has to be fixed in the same PR.

## Why

These four rules (plus complexity) were sitting in ruff's ignore list with no enforcement, so violations accumulated invisibly: exceptions were losing their root cause on re-raise, function params were shadowing Python builtins, and a real SQL-injection scanner was off repo-wide with no narrower suppression in its place. A static per-file exemption for the complexity backlog would have the same problem in miniature — it hides a PR that makes an already-flagged function even worse. The touch-to-fix ratchet is the fix for that: new debt is never grandfathered, and existing debt gets paid down as PRs naturally touch those files, without forcing a multi-week refactor up front.

### API

- Fixed all 93 pre-existing `A001`/`A002`/`B904`/`S608` violations across ~55 files: builtin-shadowing params renamed (with call sites and, where the name is a public API param like `type`, wire-compatible aliases preserved); `except` blocks now chain (`raise ... from err`) or explicitly suppress (`from None`) instead of silently dropping the original exception; the two `S608` hits were real false positives (an LLM prompt string, a test's f-string table name) and now carry ruff-specific `# noqa` justifications instead of a bandit-only `# nosec` that ruff never honored.
- Three call sites (`noop_embedding.py`, `dynamic_tool_node.py`) keep a builtin-shadowing param name because it's dictated by Chroma's/LangGraph's own call signature — verified against the installed library source and documented via `per-file-ignores`, not renamed.
- Complexity thresholds tightened (`max-args=6`, `max-branches=10`, `max-statements=50`) and enforced going forward via the new touch-to-fix ratchet described below.

### Infra

- `pyproject.toml`: `A001`/`A002`/`B904`/`S608` removed from the global ruff ignore list entirely. `PLR0911`/`PLR0912`/`PLR0913`/`PLR0915` stay listed there (so the main ruff lane doesn't double-report a debt the new script already owns) but are fully enforced through the new mechanism, not actually off.
- New `tools/lints/check_plr_complexity.py` + checked-in baseline (`tools/lints/plr_complexity_baseline.txt`): computes the full-repo violation set, diffs it against the baseline, and fails on anything genuinely new anywhere, or anything baseline-listed whose file the current PR touches. Wired into `.github/workflows/code-quality.yml` as a step in the existing Python ruff lane.
- `tools/lints/ignore_ratchet_baseline.txt` / `config/suppressions-baseline.json` updated to match — both are one-way ratchets on ruff's escape hatches and inline suppression comments, so their checked-in baselines have to move in the same commit as the config that changes them.

## How to verify

- `./.venv/bin/ruff check .` and `./.venv/bin/ruff format --check .` from repo root — both clean.
- `uv run mypy app --ignore-missing-imports` from `apps/api` — clean on all 849 source files.
- `python3 tools/lints/check_ignore_ratchet.py` and `python3 tools/lints/check_suppressions.py` from repo root — both report no new escape hatches/suppressions since their baselines.
- `python3 tools/lints/check_plr_complexity.py` from repo root — this is expected to **fail on this PR**: 17 of the files touched here for the `B904`/`A002` fixes are also on the complexity grandfather list (e.g. `create_agent.py`, `calendar_service.py`, `support_service.py`), so touching them now requires fixing their complexity too. That follow-up work is intentionally not included in this PR.

**Not verified:** the complexity refactor for those 17 touched-but-grandfathered files — tracked as explicit follow-up, not silently deferred.

## Risk

Low for the four fully-fixed rules — all fixes are mechanical (add `from err`, rename a shadowing param) and verified with the full mypy pass plus the existing test suite (12,653 passed; one pre-existing, unrelated hypothesis/DST test failure confirmed independent of this change). The complexity ratchet's risk is scoping, not correctness: any future PR that merely touches a grandfathered file now inherits an obligation to fix that file's complexity, which can surprise a contributor mid-PR until the team is used to it.


aryanranderiya (changes requested):
Blocking this until #1066 merges — it lands first and this PR rebases onto it. Flagging the overlap so the rebase is cheap:

**What #1066 changes under you**

1. **pyproject `[tool.ruff.lint] ignore`** — every entry now carries an enforced why-comment (`tools/lints/check_ignore_whys.py`). Your A001/A002/B904/S608 deletions conflict textually with those annotations; resolution is simply *your deletion wins* — removals always pass.
2. **Inline suppressions need same-line reasons** — `check_suppressions.py` fails CI on any bare `# noqa` / `# type: ignore` / `// biome-ignore`. Your ~90-file refactor must not introduce bare directives; append a short root-cause after the code if you suppress anything.
3. **File-level kill switches are governed too** — `# mypy: ignore-errors` / `# ruff: noqa` at a file top now require a reason beside them.
4. **Changed lines hit diff-coverage + mutation gates** — annotation-style edits on untested paths fail these (we learned the hard way); keep your refactor's changed lines covered by the tests you're touching.

**Governance convergence request**

Your new `plr_complexity_baseline.txt` reintroduces the checked-in-baseline pattern #1066 just deleted for suppressions (unexplained stock, no staleness detection). Not asking you to redesign here — proposing a fast-follow after both merge: migrate baseline entries into the standard model (inline `# noqa: PLR09xx -- <why>` where feasible, per-file-ignores + why-comment otherwise), then retire the txt so complexity exemptions inherit reason-enforcement and staleness detection like everything else.

**Sequencing**: #1066 merges → you rebase this; ignore-list conflicts resolve toward your deletions, everything else should be mechanical given your diffs are mostly code vs our comment-only edits.

Happy to help with the rebase when it lands.

Issue #1063, opened 2026-08-21
Python lint debt: 67 ruff@0.16 findings invisible to CI (unpinned nx lint command + no pre-commit ruff + CI pinned to 0.14.13)


aryanranderiya:
# Python lint debt: 67 ruff@0.16 findings on master that CI cannot see

Measured at `origin/master` (`dd6b19d07`), `cd apps/api && uvx ruff@0.16.4 check .` → **67 errors**.
The exact same tree under the CI-pinned `ruff@0.14.13` → **0 errors**. Nothing is wrong with the code *at 0.14.13*; every finding is a rule that was added or promoted out of preview between 0.14 and 0.16 inside rule families this repo already selects (`PL`, `FURB`, `ASYNC`, `ISC`). Because the surfaces below disagree on which ruff runs, master stays green while anyone running a current ruff sees 67 failures.

## Why CI never caught this

| Surface | Ruff version | Result |
|---|---|---|
| CI: `.github/workflows/code-quality.yml:352-358` | pinned `uvx --no-build ruff@0.14.13` | 0 errors → green |
| `nx lint api` / `mise run lint`: `apps/api/project.json:62` | **unpinned** `uvx ruff check .` → floats to 0.16.4 | 67 errors |
| Pre-commit (`.pre-commit-config.yaml`) | **no ruff hook at all** | commits never gate Python lint |

Four compounding gaps:

1. **CI pins an old ruff; local commands float to the newest.** The pin in code-quality.yml even carries a warning comment ("Unpinned, a ruff release can flip this lane red") and notes "0.16.0 shipped ISC004" — but only CI got the pin.
2. **Nobody invokes the repo's own lint target in CI.** `main.yml` deliberately doesn't duplicate static checks ("they live in code-quality.yml"), and code-quality.yml scans with its own pinned invocation — so the unpinned `nx lint api` command can rot forever without any signal.
3. **Pre-commit has no ruff hook**, so nothing gates individual commits either; debt accumulates invisibly until a rule promotion flips everything at once.
4. **The version is not pinned anywhere as a single source of truth**: `ruff>=0.4.4` in both Python projects' dev groups (root `uv.lock` resolves 0.14.13), no entry in `mise.toml [tools]`, and four separate files invoke `uvx ruff` bare (`apps/api/project.json` ×4 targets, `apps/voice-agent/project.json` ×2, `apps/voice-agent/mise.toml` ×2, root `package.json quality:py:ruff`).

[...]

PR #1065, opened 2026-08-21
fix(api): compact tool outputs into an LLM digest, JuiceFS link optional (#916)

### LINKED ISSUE #916: Compaction silently no-ops without JuiceFS; context grows unbounded
## Symptom
On any deployment without a JuiceFS workspace (all native dev), `compaction.py` caught `JuiceFSUnavailable`, logged a warning, and returned None — **no fallback**. Oversized tool outputs stayed in context, which then grew unbounded: measured 131k median input tokens per eval case, one thread growing 18,184→96,673 tokens over 14 calls, 87 'Compaction skipped' warnings in one day's log, and long tasks hitting the 100-step recursion limit.

The sibling `summarization.py` already had the correct shape: archive to workspace, catch the failure, and still do the real work.

## Fix
Committed on `feat/evals` (133d0a26d): tiered compaction — workspace spill (lossless, unchanged) when available, deterministic head+tail truncation with an honest marker when not. No LLM-summary tier by design (deterministic, free, eval-friendly; whole-history summarization stays in summarization.py). Real-storage test: 530,890 → 4,418 chars for one oversized observation.

## Residual
`insert_note` ordering (notes_utils) was a latent sibling made reachable by the Chroma fix — handled in 9b904b777 (note persists, indexing failure logs loudly and flags needs_reindex, `reindex_note` mirrors `reindex_file`).

Found by the eval harness (PR #907 work): the 'cumulative token bug' in capability runs turned out to be REAL context growth from this.

aryanranderiya:
## Summary

Compaction no longer hands the agent a file pointer and hopes it explores: an oversized tool output is digested by a model call into a dense factual summary placed directly in the conversation, and the raw output is still saved to the workspace when one exists as an optional way to recover detail. Environments without a workspace (all native development) get the summary alone instead of a lossy truncation.

## Why

On any deployment without a JuiceFS workspace, `compaction.py` had no fallback tier that preserved substance: it either spilled to the workspace or cut the output to head+tail, so a native run kept 131k median input tokens per eval case, one thread grew from 18k to 96k tokens over 14 calls, and logs showed 87 "compaction skipped" warnings in a day (#916). Even where the workspace existed, the compacted message was a heuristic preview (first three JSON items) plus a path — the agent had to spend tool calls re-mining what it had already been shown.

### API

- `compact_tool_output` composes two independent best-effort steps: the raw output is written to `/workspace/sessions/<conv>/tool_outputs/` when storage is available, and a single bounded model call digests the output into the in-context payload (input sampled head+tail, one attempt, 30s timeout, hard-capped at 4k characters). Any failure degrades to the previous deterministic tiers unchanged.
- The digest model is the graph's own chat LLM, invoked with the request's configurable bound — there is no separate model lane; it follows whatever the conversation selected (`DEV_DEFAULT_MODEL` in dev, the header selector in production).
- `init_gemini_llm`/`init_openrouter_llm` now pin their context-window profile like every other provider; fractional-token middleware raises at graph build without it.
- New compaction strategies in message metadata: `llm_summary`, `llm_summary_workspace_spill`; `workspace_spill` and `in_context_truncation` remain as degradation tiers.
- Lint commands are lockfile-pinned: nx/mise targets run `uv run --group dev ruff` (matching CI's pinned version), pre-commit gained `ruff-check`/`ruff-format` hooks, and root scripts use the same resolution (#1063).

### Infra

- `.claude/settings.json` autofix hook uses the same pinned ruff resolution.

---

## The bug

**Symptom** — On master, a deployment without the workspace mount logged `Workspace unavailable` and left oversized outputs untouched; with a workspace, the agent received `[Full output (427 KB) stored at: …]` plus a two-line preview.

**Reproduce** — Run the API natively (`ENV=development`, no mount), trigger any tool returning >120k characters, and inspect the ToolMessage that replaces it: head+tail truncation on master; a digest on this branch.

**Root cause** — `compact_tool_output` treated the spill as the only real tier: its success path built the replacement message from `_summarize_output` (first three list items / first 500 characters) plus a path, and every failure path returned truncation (`apps/api/app/agents/middleware/compaction.py`). No model was ever asked to compress meaning.

**The fix** — The spill became optional and a bounded model digest became the primary payload, with the deterministic tiers kept as literal fallbacks. Because the digest rides the graph's chat LLM, dev environments with only `DEV_LLM_*` configured get working compaction automatically instead of silently losing the tier.

**Regression test** — `tests/unit/agents/test_compaction_middleware.py::TestLLMSummary::test_digest_stands_alone_without_a_workspace` fails on master (output truncated) and passes here (summary carried, strategy `llm_summary`). Seen red before the fix landed.

**Why the suite missed it** — The old tests asserted the spill contract itself (pointer present, file written) — they encoded the behavior being fixed, so nothing flagged the missing-substance case.

---

## How to verify

1. Run the API natively with no JuiceFS mount and `DEV_LLM_*` configured.
2. Send a chat message that triggers deep research; expect the log line `Compacted tool output into an LLM summary` and a final answer whose specifics (counts, names, figures) match the sources.
3. Ask a detail question about the researched material in a follow-up turn; expect it answered from context without re-running the tool.
4. Stop the workspace (or run outside Docker) and repeat step 2; expect the same digest quality with no spill warning beyond `Workspace unavailable`.

**Not verified:** a Docker'd stack with a live JuiceFS mount — the spill path is exercised at the `write_session_file` seam in tests only.

| Metric | Before | After | How measured |
| --- | --- | --- | --- |
| Compacted payload for a 427 KB tool output | ~500-char preview + path | ~1.4k-char digest | checkpoint inspection |
| Follow-up turn needing re-exploration | every time | not required | live chat drive |

Figures come from a manual live session against deepseek-v4-flash; the measurement script was throwaway.

## Risk

The digest call adds latency inside the tool loop for exactly the outputs that were already oversized (single attempt, 30s cap, then deterministic degradation). If the configured endpoint is slow, users see slower tool turns rather than failures; if it is down entirely, compaction behaves as it did on master.

## Related

Fixes #916 · Refs #1063


aryanranderiya:
Addressing the review + live-drive request:

**1. Dev-LLM routing (fixed in cd48dd37):** `get_summarization_llm()` is back to master semantics — default model only, dropped when unconfigured. Whole-history summarization never touches the dev lane. A new `get_compaction_summary_llm()` is the *only* consumer allowed onto the DEV_LLM_* endpoint (development-only, custom provider registered), serving tool-output digests; 8 new tests pin this boundary.

**2. Profile exposure:** verified empirically rather than assumed — the custom provider's configurable wrapper resolves `.profile` → `{'max_input_tokens': 1000000}`, and `SummarizationMiddleware(model=<wrapper>, trigger=("fraction", …))` constructs cleanly. After the routing split, history summarization never receives the wrapper regardless.

**3. Live full-stack drive** (real API + Postgres/Mongo/Redis, dev-bypass auth, `OPENROUTER_API_KEY` stripped so chat rode `DEV_DEFAULT_MODEL=custom` → opencode-gateway deepseek-v4-flash and digests exercised the true DEV_LLM fallback):
- Turn 1 (exhaustive depth-3 research): log chain was exactly `compaction digests use the DEV_LLM_* custom endpoint` → `Workspace unavailable, compacting without a spill` → **`Compacted tool output into an LLM summary` ×2**. Executor delivered a 13,274-char report with specifics intact (Samsung SDI 500 Wh/kg / 9-min charge / 20-yr life; QuantumScape 1,000 cycles @ >95% retention).
- Turn 2 (detail probe: "what charge time/lifespan did Samsung promise? which automaker evaluates them?"): answered **entirely from the compacted digests — zero re-research calls** — every figure correct.

Full suite: 11,826 passed / lint+type-check green.

aryanranderiya:
Follow-up (8f20112f8) — you were right, the accessor functions shouldn't exist:

**One LLM now.** `get_summarization_llm` / `get_compaction_summary_llm` / `_get_dev_custom_llm` are deleted. The middleware receives the graph's own `chat_llm` — the same configurable instance the chat flow runs — and invokes it inside the graph where ambient request config routes it to exactly the model the conversation uses (`llm.with_config(configurable=...)`, same as the model node). Dev with only `DEV_LLM_*`: everything rides that endpoint automatically. Production: everything rides the request-pinned model.

**Found + fixed while re-driving live:** with OpenRouter stripped, chat fell to Gemini and the comms graph died — `init_gemini_llm`/`init_openrouter_llm` never pinned the context-window profile that fractional-token middleware requires (`init_custom_llm`/`_build_default_llm`/`_sim_llm` did). All chat LLMs now pin it; the contract is uniform.

**Verified:** full unit suite 11,825 passed incl. 7 new wiring tests (middleware gets `chat_llm`; digest call binds request configurable; no-llm keeps deterministic tiers); lint+type-check green; live API drive on the new wiring — graphs build, streaming works, tools dispatch (the earlier hard comms failure is gone). The deep-research→compaction cycle itself was already demonstrated end-to-end in the previous round's drive (13k-char report + digest recall); this round's repeat run hit this sandbox user's spend guards mid-research (uncatalogued Gemini priced at DEFAULT_PRICING inflates the accounting budget), so the routing proof rests on the wiring tests plus successful graph runs rather than a second paid research cycle.

PR #1067, opened 2026-08-21
fix(workflows): JSON-safe tool payloads, reschedule-safe scheduled fires, break-sentinel variants

Dhruv-Maradiya:
## Summary

Fixes three workflow bugs in the agent tool-calling path: creating or editing a scheduled workflow no longer dies with "Object of type datetime is not JSON serializable", rescheduling a workflow now actually moves its next run (it used to keep firing at the original time), and Telegram messages no longer show literal `<NEW_LINE_BREAK>` text when the model emits a near-miss spelling of the bubble-break sentinel.

## Why

All three were reported from real usage. The datetime crash wedged the agent in a retry loop exactly when a user asked GAIA to create or change a schedule over chat; the reschedule bug silently ran workflows at times the user had already changed; and the stray `<NEW_LINE_BREAK>` tokens made bot replies look broken.

### API
- `get_workflow`, `pause_workflow`, `resume_workflow` and `apply_workflow_edit` now serialize workflow payloads with `model_dump(mode="json")` — these are handed to both the LLM as a tool result and the SSE stream writer, and both plain `json.dumps` them.
- Scheduler-originated ARQ jobs carry a `scheduled_for` stamp (the occurrence they were armed for), and `claim_for_execution` pins `trigger_config.next_run` in Mongo, so a deferred fire whose occurrence was rescheduled away is rejected atomically instead of executing. Jobs enqueued before this change carry no stamp and behave exactly as before, so a deploy never strands an existing schedule.
- Workflow-result delivery to messaging platforms splits bubbles on any spelling of the break sentinel (`<NEW_MESSAGE_BREAK>`, `<NEW_LINE_BREAK>`, case/whitespace variants).

### Bots / Mobile
- Shared streamer (`libs/shared/ts`) normalizes break-sentinel variants to the canonical token before splitting bubbles, previews, and sealing the live bubble, and recognizes partial variants split across stream chunks — so no spelling ever reaches a platform as literal text.

## The bug

**Symptom** — Three symptoms, one user journey ("create a workflow for 4pm" → "actually make it 9pm"):
1. `TypeError: Object of type datetime is not JSON serializable` surfaced as a tool error and the agent retried endlessly.
2. A workflow edited from 16:00 to 21:00 still ran at 16:00.
3. Telegram replies contained literal `<NEW_LINE_BREAK>` strings.

**Reproduce** — On master, over chat: ask GAIA to create a daily 16:00 workflow, then edit it to 21:00. The edit path streams `updated.model_dump()` through `make_redis_stream_writer` (`redis_writer.py:42`), which raises inside the tool; and after the edit, the original 16:00 ARQ job still fires and passes the claim gate because `activated=True` and `status="scheduled"` both hold.

**Root cause** —
1. Python-mode Pydantic dumps keep native datetimes; every consumer of these payloads serializes with plain `json.dumps`.
2. ARQ has no job cancellation. The claim gate verified liveness and run-state but not *which occurrence* the fire was armed for, so a stale deferred job was indistinguishable from a live one.
3. Every splitter matched only the exact canonical token `<NEW_MESSAGE_BREAK>`.

**The fix** — Serialize at the boundary (`mode="json"`); make the workflow document the source of truth at fire time by pinning the claimed occurrence against `trigger_config.next_run`; normalize sentinel spellings before splitting. No ARQ deletion or abort machinery — worst case a stale fire is skipped, never wrongly executed.

**Regression test** —
- `tests/unit/tools/test_workflow_tool_json_safety.py` (bug 1) and `tests/unit/services/test_workflow_scheduler_reschedule.py` + worker gate tests (bug 2): all seen failing on base before the fix.
- Contract tests against real Mongo in `tests/contracts/test_workflows_repository.py`: stale-fire rejection and JSON-safe payload round-trip on persisted documents.
- Bot streaming variant tests in `apps/bots/__tests__/shared/utils/streaming.test.ts`.

**Why the suite missed it** — No test exercised the workflow tools' stream frames or tool results with a real model instance (mocks returned attribute bags without datetimes), nothing covered the interaction between deferred ARQ jobs and a mid-life reschedule, and the bot splitter tests only ever used the canonical token.

## How to verify

On a booted stack (`mise dev --sim` + `nx worker api`):
1. Chat: "create a workflow that runs every day at 4pm" → workflow is created, activated, and an ARQ job exists for 16:00.
2. Chat: "reschedule it to 9pm" → Mongo shows `next_run` at 21:00 and a second ARQ job appears.
3. Force-fire the old job (`zadd arq:queue <now_ms> execute_workflow_by_id:<id>:<old_ts>`) → the worker skips it ("skipped duplicate scheduled fire"), zero executions recorded.
4. Force-fire the new job → it passes the gate and runs.

Verified end to end this way on a live stack (real executor/subagent LLM path, real Mongo/Redis/worker). Not exercised: a real Telegram connection — bot-side behavior is proven through the shared-streamer test suite and the platform-emulating harness.

## Risk

Low-to-moderate. The one behavioral tightening: a scheduler fire whose stamp doesn't match `trigger_config.next_run` is now skipped. Every current writer (create, re-arm, activate, reap, reschedule) keeps those two values in lockstep, and unstamped legacy jobs are ungated — but a hypothetical row where they diverge would skip fires rather than run at a stale time, which is the intended direction. Deploy-safe: in-flight jobs enqueued before this change carry no stamp.


apps/api/app/agents/core/background/workflow_platform_delivery.py
```diff
@@ -31,6 +31,11 @@
 from app.utils.notification.channel_preferences import fetch_channel_preferences
 from shared.py.wide_events import log
 
+# The model occasionally emits a near-miss spelling of the bubble-break sentinel
+# (e.g. <NEW_LINE_BREAK> instead of <NEW_MESSAGE_BREAK>). Every spelling splits
+# bubbles; none ever ships to a platform as literal text.
```
Dhruv-Maradiya: constants file

apps/api/app/agents/core/background/workflow_platform_delivery.py
```diff
@@ -57,7 +62,7 @@ async def deliver_workflow_result_to_platforms(
 
     # Comms splits its reply into bubbles with the break sentinel;
     # publish_outbound_message strips blanks and sends them as one ordered message.
-    bubbles = notification_text.split(NEW_MESSAGE_BREAKER)
+    bubbles = _BREAK_SENTINEL_RE.split(notification_text)
```
Dhruv-Maradiya: not sure if this solution will work I suspect it could be bcz of streaming in telegram

Dhruv-Maradiya:
Addressed all review comments in the latest commits:

**@Dhruv-Maradiya**
- **constants file**: the break-sentinel regex moved to `app/constants/general.py` as `MESSAGE_BREAK_SENTINEL_RE`, next to `NEW_MESSAGE_BREAKER` — one place now owns the sentinel vocabulary.
- **streaming concern**: your instinct was right that Telegram streaming is the path where you saw the leak — and it's covered. Live chat goes through the shared streamer (`libs/shared/ts/src/bots/utils/streaming.ts`), which normalizes variant spellings (any casing, optional whitespace) before splitting bubbles and recognizes *partial* variants split across stream chunks, with tests emulating exactly that chunking (`apps/bots/__tests__/shared/utils/streaming.test.ts`). The Python change covers the other path that reaches Telegram: background workflow results delivered via the outbound queue (no streaming there). Both paths are pinned by tests.

**CodeRabbit**
- `noqa: ARG002` → parameter renamed to `_scheduled_at` (done before review landed).
- Stamp/claim precision: `TriggerConfig.calculate_next_run` now returns whole seconds only, so the ARQ stamp round-trip can never diverge from the stored occurrence regardless of sub-second components.
- Side-effect import: removed entirely; test helpers typed (`RunnableConfig` / `object`).
- Partial-token grammar mismatch: fixed — `stripPartialBreakToken` now matches partial variants case-insensitively with optional whitespace, same grammar as normalization.

**Greptile**
- P1 partial variants: same fix as above, plus new streaming tests for lowercase/spaced partials split mid-chunk.
- P1 untrusted `scheduled_for`: non-numeric stamps are now treated as unstamped (ungated, pre-change behavior) and logged instead of crashing `fromtimestamp`; numeric-but-unrepresentable values (overflow) take the same discard path. Both pinned by regression tests.

**Sonar S1192**: `NEXT_RUN_FIELD` constant extracted in the workflows repository.

All 9 touched modules re-ran through the mutation gate at zero survivors on changed lines.

Dhruv-Maradiya:
All review feedback resolved on head (69963afe7):

**@Dhruv-Maradiya**
- **constants file**: break-sentinel regex moved to `app/constants/general.py` as `MESSAGE_BREAK_SENTINEL_RE`, next to `NEW_MESSAGE_BREAKER`.
- **streaming concern**: confirmed covered — live Telegram chat goes through the shared streamer (`streaming.ts`), which normalizes variant spellings before splitting and recognizes partials split across chunks (tests emulate exactly that chunking). The Python change covers the other path reaching Telegram: background workflow results via the outbound queue (no streaming there).

**CodeRabbit**
- `noqa: ARG002` → renamed to `_scheduled_at` (suppression-baseline lane green).
- Stamp/claim precision → `calculate_next_run` returns whole seconds; stamp round-trip can't diverge.
- Test `Any` types → `RunnableConfig` / `object`.
- Side-effect import → removed (cycle no longer exists); remaining imports at module scope.
- Partial-token grammar → `stripPartialBreakToken` now case-insensitive + whitespace-tolerant, matching normalization.

**Greptile**
- P1 partial variants → same fix, plus mid-chunk variant tests.
- P1 untrusted `scheduled_for` → non-numeric stamps treated as unstamped + logged instead of crashing; numeric overflow takes the same path. Both pinned by regression tests.

**Sonar S1192** → `NEXT_RUN_FIELD` constant extracted.

Note: @Dhruv-Maradiya your review on this PR is still *pending*, which blocks inline thread replies — submit it when ready and the threads can be marked resolved.

Dhruv-Maradiya (commented):
All feedback addressed on 69963afe7 — sentinel regex moved to constants, streaming path covered and tested.

apps/api/app/services/workflow/scheduler.py
```diff
@@ -61,13 +61,28 @@ def get_job_name(self) -> str:
...
+        the worker compares the stamp against the workflow's current
+        ``trigger_config.next_run`` and skips the stale fire instead of running
+        the workflow at its original time.
+        """
+        return (
+            task_id,
+            {
+                "trigger_type": TriggerType.SCHEDULE.value,
+                "scheduled_for": int(scheduled_at.timestamp()),
+            },
```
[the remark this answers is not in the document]
Dhruv-Maradiya: Resolved: `TriggerConfig.calculate_next_run` now returns whole seconds only (`.replace(microsecond=0)`), so every armed occurrence is exactly representable in BSON and the stamp round-trip can never diverge from the stored `next_run`. The sub-second scenario can no longer occur at the source.

apps/api/app/workers/tasks/workflow_tasks.py
```diff
@@ -521,15 +521,32 @@ async def execute_workflow_by_id(
...
+        # cancellation, so after a reschedule the old deferred job still fires;
+        # trigger_config.next_run has moved on and the mismatch rejects it instead
+        # of running the workflow at its original time. Jobs enqueued before the
+        # stamp existed carry no key and are ungated, so a deploy never strands a
+        # schedule.
+        if trigger_type == TriggerType.SCHEDULE.value:
+            scheduled_for = context.get("scheduled_for") if context else None
+            expected_next_run = (
+                datetime.fromtimestamp(scheduled_for, tz=UTC) if scheduled_for is not None else None
             )
```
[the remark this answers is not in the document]
Dhruv-Maradiya: Resolved: only a numeric stamp is treated as scheduler provenance now. Non-numeric values (a manual caller typing its own context) and numeric-but-unrepresentable ones (fromtimestamp overflow) are discarded as unstamped — ungated, pre-change behavior — with a warning carrying workflow_id and the truncated value. Pinned by regression tests (garbage string + 10**40 overflow), all verified through the mutation gate at zero survivors.

apps/api/app/agents/core/background/workflow_platform_delivery.py
Dhruv-Maradiya: Resolved: moved to `app/constants/general.py` as `MESSAGE_BREAK_SENTINEL_RE`, right beside `NEW_MESSAGE_BREAKER` — one place owns the sentinel vocabulary.

apps/api/app/agents/core/background/workflow_platform_delivery.py
Dhruv-Maradiya: Your suspicion was correct about where you saw it: live Telegram chat goes through the shared streamer, so that's fixed in `libs/shared/ts/src/utils/messageBreakUtils.ts` + `streaming.ts` — variant spellings (any casing/whitespace) normalized before bubbles split, partial tokens split across stream chunks also handled. The Python change here covers the other path that reaches Telegram: background workflow results via the outbound queue (no streaming on that path). Both paths pinned by tests: emulated Telegram chunked-variant cases in `apps/bots/__tests__/shared/utils/streaming.test.ts` and bubble-split tests in `test_workflow_platform_delivery.py`, plus a live end-to-end run through the real bot pipeline.

PR #1083, opened 2026-08-23
fix(api): drop the unreachable str branch in the chroma inventory

aryanranderiya:
## What

`_chroma_inventory` in the account-deletion script narrowed a collection with
`isinstance(col, str)`. On the pinned chromadb (1.0.13),
`ClientAPI.list_collections()` is typed `Sequence[Collection]`, so that branch can
never be taken and mypy rejects it as `[unreachable]`.

```
apps/api/app/scripts/delete_user_account.py:138:34: error: Subclass of "Collection"
and "str" cannot exist: would have incompatible method signatures  [unreachable]
```

## Why now

**`master` is currently red on the `python-mypy` lane and has been since #1081.**
That PR was merged with a failing Quality gate, and every master run since has been
web-only — the lane self-skips when no `.py` files change, so the failure stopped
being visible.

It is not harmless: the lane checks *project roots*, not changed files
("mypy needs the whole module graph per project, so it can't be scoped"), so **any**
PR touching `apps/api/app` inherits this failure. Two other fixes I have open are
blocked behind it.

## Verification

Ran the exact CI invocation on a clean, CI-identical venv
(`uv sync --project apps/api --group backend --group dev`):

```
uv run --project apps/api --frozen --group backend --group dev \
  mypy apps/api/app apps/voice-agent/src libs/shared/py tools/lints --ignore-missing-imports
```

- before: `Found 1 error in 1 file (checked 888 source files)`
- after: `Success: no issues found in 888 source files`

The `repository_boundaries` lint on this file also passes (`scripts/` is already in
`_BSON_ALLOWED_DIRS`, so the `bson` import flagged in #1081's run is resolved).

## Note

No test accompanies this one: it deletes an unreachable branch, so there is no
behaviour to pin — mypy is the regression check, and it is an enforced CI lane.

### DIFF
--- a/apps/api/app/scripts/delete_user_account.py
+++ b/apps/api/app/scripts/delete_user_account.py
@@ -135,7 +135,7 @@ def _pg_inventory(conn: psycopg.Connection[Any], uid: str) -> dict[str, int]:
 def _chroma_inventory(client: chromadb.api.ClientAPI, uid: str) -> dict[str, int]:
     counts: dict[str, int] = {}
     for col in client.list_collections():
-        name = col if isinstance(col, str) else col.name
+        name = col.name
         got = client.get_collection(name).get(where={"user_id": uid}, limit=200_000, include=[])
         n = len(got.get("ids") or [])
         if n:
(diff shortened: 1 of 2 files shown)


aryanranderiya:
Folded into #1085 and #1084 as cherry-picked commits at the author's request, so each is green standalone — whichever merges first makes the other copies no-ops. Closing as redundant.

PR #1084, opened 2026-08-23
fix(payments): accept Dodo payment webhooks whose tax and method are null

aryanranderiya:
## The bug

Dodo sends `tax`, `settlement_tax` and `payment_method` as `null` when a payment
never settles and no method is captured — a declined authorization or an abandoned
checkout. `DodoPaymentData` modelled all three as required `int`/`str`, so Pydantic
rejected the payload:

```
[PAYMENT] Dodo payment webhook payload did not validate
3 validation errors for DodoPaymentData
tax             Input should be a valid integer  [input_value=None]
settlement_tax  Input should be a valid integer  [input_value=None]
payment_method  Input should be a valid string   [input_value=None]
```

Seen in production on `api.heygaia.io`, immediately after a **successful** signature
verification — so this is a genuine Dodo webhook we reject.

## Why it matters — the event is acked and lost

The failure is silent by construction:

1. `get_payment_data` logs and returns `None`.
2. Each handler turns that into `ValueError("Invalid payment data")`.
3. `process_webhook` catches it and returns `status="failed"`.
4. The endpoint answers **HTTP 200** — deliberate, so Dodo's retry policy follows the
   processing result rather than a request rejection.

Net effect: Dodo sees a 200 and **never retries**, `_mark_webhook_as_processed` is
never reached, and the event is dropped. No failure analytics, no dunning signal, and
a single ERROR line as the only evidence. Recovery is manual.

## The fix

Make the three fields `| None = None`. **None of them is read anywhere in the
codebase** (`grep` for `.tax` / `settlement_tax` / `payment_method` across
`apps/api/app` returns only their own declarations) — they were required but unused,
so their strictness bought nothing and cost us dropped webhooks. Every genuinely
nullable neighbour (`card_network`, `card_last_four`, `subscription_id`, …) was
already modelled this way; these three were the outliers.

## Test

`test_processes_a_failure_whose_tax_and_method_are_null` drives a `payment.failed`
event with all three null through `process_webhook`.

**Watched it fail first**, reproducing the exact production log line:

```
E  AssertionError: assert 'failed' == 'processed'
   [PAYMENT] Dodo payment webhook payload did not validate  (webhook_models.py:124)
```

Then green: `88 passed`.

The gap this closes: `PAYMENT_DATA_PAYLOAD` is a single fully-populated fixture reused
for *every* payment event type including `payment.failed`, so the real shape of a
failed payment was never exercised. `test_invalid_payment_data_raises` covers a
wholly-missing payload (`{"incomplete": True}`), not present-but-null fields.

## CI

The pre-existing `python-mypy` failure on `master` (an unrelated `[unreachable]`
error in `app/scripts/delete_user_account.py`, merged red in #1081) is **folded into
this PR** as two cherry-picked commits — the one-line fix plus the tests the mutation
gate requires for that module — so this PR is green standalone. The same commits ride
on the sibling PR; whichever merges first makes the other's copy a no-op.

### DIFF
--- a/apps/api/app/models/webhook_models.py
+++ b/apps/api/app/models/webhook_models.py
@@ -61,10 +61,14 @@ class DodoPaymentData(BaseModel):
     total_amount: int
     settlement_amount: int
     settlement_currency: str
-    tax: int
-    settlement_tax: int
+    # Null whenever the payment never settled or no method was captured — a declined
+    # authorization or an abandoned checkout arrives with all three unset. Modelling
+    # them as required rejected those payloads, and since the endpoint answers 200 by
+    # design the event was acked and lost rather than retried.
+    tax: int | None = None
+    settlement_tax: int | None = None
     status: str
-    payment_method: str
+    payment_method: str | None = None
     card_network: str | None = None
     card_type: str | None = None
     card_last_four: str | None = None
--- a/apps/api/app/scripts/delete_user_account.py
+++ b/apps/api/app/scripts/delete_user_account.py
@@ -135,7 +135,7 @@ def _pg_inventory(conn: psycopg.Connection[Any], uid: str) -> dict[str, int]:
 def _chroma_inventory(client: chromadb.api.ClientAPI, uid: str) -> dict[str, int]:
     counts: dict[str, int] = {}
     for col in client.list_collections():
-        name = col if isinstance(col, str) else col.name
+        name = col.name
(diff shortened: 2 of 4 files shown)


aryanranderiya:
Closing per the author: the null-field payload came from a manual plan update (no charge, discount removed) rather than an organic flow, and the dropped-event risk is accepted. The webhook will continue to log a validation ERROR and ack such payloads without processing them; reopen if failed-payment analytics/dunning signals become wanted.

PR #1170, opened 2026-08-31
fix(api): anchor the llm_calls backfill to OpenRouter's own billed totals

aryanranderiya:
## What

The history backfill (#1157's `scripts/backfill_llm_calls.py`) under-reported real OpenRouter spend by a third: $53.40 computed vs $80.05 billed for Aug 11–30. This PR makes the backfill anchor on OpenRouter's own books so an anchored day's ledger total equals OpenRouter's billed total for that day, exactly.

## Why the old numbers were wrong — measured, not guessed

Reconciled the log events against `GET /api/v1/activity` (per day × model × provider: real dollars, real token counts, real request counts):

- **Stable days (Aug 19–23, 25–29): requests within 1–4%, prompt tokens within 1–8%** — the events are nearly complete, so the dollar gap on those days is pure pricing: the flat table can't know which upstream served a call and real rates spread 9× (Baidu $0.045/M vs Reka/Wafer ~$0.20/M effective).
- **Aug 13–18: the metering seam logged many calls twice** (copies differ only in timestamp microseconds; sometimes only one carries the generation id). After collapsing them, Aug 17 = 1,673 ours vs 1,680 billed.
- **Aug 6–12 and Aug 24: OpenRouter billed thousands of requests with few or no usable events** (pre-cost-field era, log retention, deploy-day chaos) — ~$11 of real spend with nothing to attribute it to.
- Old events also never counted reasoning tokens, which OpenRouter bills as output.

## How

- **`scripts/_activity.py`** — loads `--or-activity` snapshot files (the window is rolling ~25 days, so snapshots preserve days the API will refuse to return next week) and/or fetches live via `OPENROUTER_MANAGEMENT_KEY`; merges them per (day, model, provider, endpoint), larger `usage` wins.
- **Allocation pass** — per anchored (day, model): the pool's real dollars, scaled by the share of the day's tokens our events observed, are distributed across rows in proportion to each row's best-known cost (generation lookups and provider-captured costs act as weights, so per-call price shape survives). `cost_source: "allocated"`.
- **Unattributed remainder** — what OpenRouter billed beyond what events can attribute becomes one explicit `or_unattributed` row per (day, model) (background, never charged, deterministic `backfill_key`) — visible as a gap, never smeared, never dropped. Days with no events at all (Aug 6–12) become fully unattributed rows.
- **Double-log collapse** — same second + same (user, model, token counts) is one call; the copy with a generation id is kept; two *different* generation ids are never merged. Counted in the anomalies table.
- **Live-row ceiling** — events at/after the first non-backfilled ledger row are skipped (auto-detected on `--apply`, `--until` for dry runs), so live and backfilled rows can never double-count.
- **Today's pool is never anchored** — a mid-day activity fetch is a partial total.
- `CostSource` gains `"allocated"`; repository gains `first_live_call_at()`.

## Verified

- `uv run pytest tests/unit/scripts/test_backfill_llm_calls.py tests/unit/db/test_llm_calls_repository_unit.py -q` → **49 passed** (new tests pin: anchored day sums to pool exactly; remainder explicit; over-coverage capped at pool; unanchored models untouched; orphan pools become rows; today skipped; deterministic remainder keys; double-log collapse keeps the auditable copy; distinct generation ids never merge).
- Allocation invariants additionally exercised standalone (day total == pool to 1e-9 with weights honoured).
- `ruff` + `mypy` clean on all four files; pre-commit hooks green.
- NOT yet verified: the full dry run against production Loki + the real activity snapshot — that run is the next step and its output will be posted before any `--apply`.

## Rollout

1. Dry-run in the prod api container: `python scripts/backfill_llm_calls.py --dry-run --or-activity /tmp/or_activity.json` (snapshot of Aug 6–30, $85.92, already captured).
2. Review the per-day table (`$` must equal `OR $` on every anchored day) and the anomalies table.
3. `--apply` — idempotent via `backfill_key` unique index.

### DIFF
--- a/apps/api/app/db/repositories/llm_calls.py
+++ b/apps/api/app/db/repositories/llm_calls.py
@@ -24,7 +24,7 @@ what keeps the collection bounded — the durable per-day money history stays in
 """
 
 from collections.abc import Sequence
-from datetime import datetime
+from datetime import UTC, datetime
 import re
 from typing import Literal, NamedTuple
 
@@ -34,7 +34,7 @@ from pymongo import UpdateOne
 from app.constants.general import EXECUTOR_THREAD_PREFIX, SPAWN_THREAD_PREFIX
 from app.db.repositories.base import MongoDocument, MongoRepository
 
-CostSource = Literal["provider", "table"]
+CostSource = Literal["provider", "table", "allocated"]
 CallStatus = Literal["ok", "error"]
 #: Short, stable classification of WHY a provider call failed. Derived from the
 #: exception type, never from its message: messages carry model ids, prompts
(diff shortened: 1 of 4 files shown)


aryanranderiya:
Running this as a one-off operational script copied into the prod container instead of merging — the ledger's history only needs backfilling once.

PR #1194, opened 2026-09-03
feat(analytics): attribute LLM spend to feature, surface and workflow

Vinitthakkar45:
## Summary

Every LLM call GAIA makes now says what it was *for*, where it came *from*, and which workflow asked for it. PostHog has been pricing agent calls since July, but only as an undifferentiated total — you could see that a user cost $6.51 and not whether that was chat, a scheduled workflow, or Gmail. This adds those dimensions to the event PostHog already emits, and adds one new event for the background work PostHog never saw at all.

The immediate payoff is unit economics: cost per user split by plan, and the first visibility into what GAIA spends on its own behalf versus what users ask for.

## Why

Free-tier LLM spend is $133.58/month across 359 users against $18.88 from 6 Pro users (last 30 days, measured in PostHog). Nobody could see that split before, because no event carried a plan tier or a feature.

Separately, background spend — memory extraction, onboarding inference, follow-up generation, conversation titles — reached PostHog not at all. `posthog.ai.langchain.CallbackHandler` is only attached to agent-graph runs, so those one-shots were priced into `usage_daily` and then invisible.

### API

- **`$ai_generation` now carries `feature`, `surface` and `workflow_id`.** One extra properties dict on the PostHog callback handler (`helpers/agent_helpers.py`). `feature` is chat / workflow / integration, derived from the workflow id and the agent name; `surface` is `ui` / `bot` / `bg`, read off the run's existing `source_category`. No new event on this path, so nothing is double-counted.
- **New `ai:llm_call_completed`, background calls only.** Emitted from the auxiliary metering seam in `agents/llm/client.py`, which is the only place that sees one-shots. Carries the four token counts, the cost from GAIA's own rate card, and the feature.
- **`feature` is a required argument** on `ainvoke_structured` / `ainvoke_structured_gemini`, so the type checker enumerates any call site that omits one. 21 auxiliary sites now state theirs.
- **`cost_estimated` flags an untrustworthy dollar figure.** `model_pricing.py` prices an unregistered model at a generic default rather than raising, so the cost looks plausible and is wrong by a large factor. New `has_rate_card()` makes that case visible instead of silent.
- **`rate_limit_hit` fixed** — see below.

## The bug

**Symptom** — `rate_limit_hit` events carry `plan: null`, so throttling cannot be segmented by Pro vs Free. 79,211 events over 30 days, every one unattributed to a tier.

**Reproduce** — On master: hit any tiered limit as a free user, then query PostHog for `rate_limit_hit` and read `properties.plan`. It is absent.

**Root cause** — `services/limit_upsell.py:79` emits the event with a raw `"rate_limit_hit"` string and only `{feature, origin}`. The two emitters in `decorators/rate_limiting.py` do pass `plan`, but the `limit_upsell` path handles the free-tier upsell and is the dominant emitter, so in practice `plan` was never set. The raw string is also why a search for `AnalyticsEvents.RATE_LIMIT_HIT` does not find this call site.

**The fix** — Use the enum, and pass `user_plan` through `_run` so the property is set at the emitter that actually fires.

**Regression test** — `tests/unit/services/test_limit_upsell.py` and `tests/unit/api/test_tiered_rate_limiter.py` asserted the old shape and were pinning the bug; they now assert the enum and the `plan` property. Both were seen failing against the fix before being updated.

**Why the suite missed it** — The tests asserted exactly what the code did, raw string and missing property included. Nothing compared the emitted shape against the other two emitters of the same event name.

## How to verify

1. Boot the API and send a chat message. In PostHog, open the newest `$ai_generation` and confirm `feature: chat` and `surface: ui`.
2. Run a workflow (`POST /api/v1/workflows/{id}/run`). Its `$ai_generation` events should read `feature: workflow`, `surface: bg`, and carry the `workflow_id`.
3. Trigger a Gmail or Calendar action so a subagent runs. Those calls report `feature: integration` with `agent_name` naming the integration.
4. Let memory extraction run (send two messages in one conversation). A single `ai:llm_call_completed` should appear with `feature: memory`, `surface: bg`, `charged: false`.
5. Exhaust a free-tier limit and confirm the `rate_limit_hit` event now carries `plan: free`.

**Not verified:** no bot or voice turn was exercised. Both route through the same agent graph as chat, so they should inherit the dimensions, but that is reasoned rather than observed.

## Post-merge steps

- [ ] Confirm the dashboard tiles populate — the five that read `feature` / `surface` / `workflow_id` show `unattributed (deploy pending)` until this ships: [AI Spend & Token Usage](https://us.posthog.com/project/201142/dashboard/2040295)

## Metrics

| Metric | Before | After | How measured |
| --- | --- | --- | --- |
| Cost per Pro user / month | unknown | $3.15 | PostHog, `$ai_generation` joined to persons, 30d |
| Cost per Free user / month | unknown | $0.37 | same |
| Free-tier share of LLM spend | unknown | 87% ($133.58 of $154.77) | same |
| `rate_limit_hit` events with a plan | 0 of 79,211 | all new events | PostHog, 30d |

Figures are reported from the live project, not reproducible from this diff. Background spend is not in them: it only starts flowing once this deploys.


apps/api/app/services/onboarding/first_message_service.py
```diff
@@ -106,7 +107,10 @@ async def generate_first_message(
         llm = get_helper_llm()
         t_llm = time.monotonic()
         response = await ainvoke_llm(
-            llm, [HumanMessage(content=prompt)], label="onboarding_first_message"
+            llm,
+            [HumanMessage(content=prompt)],
+            label="onboarding_first_message",
+            feature=AIFeature.ONBOARDING,
         )
```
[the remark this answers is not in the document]
Vinitthakkar45: Correct, and it was a real bug — thanks.

Fixed in `5001b0c73`: the call now passes `config=metered_config(user_id)`.

Accountability: this was not pre-existing debt I inherited, it was a hole I opened. `first_message_service` never needed a config before this PR, because before this PR nothing downstream of `ainvoke_llm` cared who the call belonged to. I added the consumer (`capture_auxiliary_llm_call`) and audited the call sites that *had* no `user_id` in scope, but not the ones that had it and simply were not forwarding it. Onboarding first-message generation is one of the highest-volume background calls we make, so it was also one of the most expensive ones to miss.

Two things came out of it beyond the one-liner:

1. Silence was the actual defect. A call with no user was being dropped without a trace, which is exactly why nobody would have noticed. `capture_auxiliary_llm_call` now emits `log.warning("llm_call_unattributed", label=..., model=...)` on that path, so the next one of these shows up as a log line naming the label instead of as a quiet gap in the dashboard.
2. That warning path is itself now mutation-tested, so a future refactor cannot silently delete it.

apps/api/app/helpers/agent_helpers.py
```diff
@@ -500,7 +506,12 @@ async def build_agent_config(
     )
 
     callbacks = _build_agent_callbacks(
-        conversation_id, user, agent_name, tracing.usage_metadata_callback
+        conversation_id,
+        user,
+        agent_name,
+        source,
+        (base_configurable or {}).get("workflow_id"),
+        tracing.usage_metadata_callback,
```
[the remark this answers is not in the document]
Vinitthakkar45: Real bug, and the diagnosis is exactly right. Fixed in `5001b0c73`.

The ordering was the whole problem: `build_agent_config` builds the callbacks, and `_core_agent_logic` stamps `workflow_id` onto the configurable *after* that call returns, so on a top-level workflow fire there was nothing to read. Rather than move the stamp (which several other readers depend on landing where it does), the turn now carries it: `AgentTurn.workflow_id`, populated from `trigger_context` at the one place that knows a workflow is firing, and `_build_agent_callbacks` reads `turn.workflow_id or inherited.get("workflow_id")` so both the top-level run and its children resolve.

Accountability: I verified this on a child run, where the inherited configurable *does* already carry `workflow_id`, and generalised from that single observation to the top-level case — which is precisely the case that had none. Reading the code made it look symmetric; it is not.

Two follow-ons worth recording, because reviewing this one properly turned up a second bug of the same shape:

1. When I reverted my fix to check the new test actually catches it, the caller-side half stayed **green**. The test was asserting the helper, not the wiring, so the gap it was written for was still open. `tests/unit/agents/test_agent_core.py` now covers that the workflow id reaches `build_agent_config` at all.
2. `source` had the identical defect one line over — `_inherit_from_parent_configurable` fills `conversation_source` from the parent, but I was passing raw `turn.source`, so every executor and subagent run (i.e. most of our token spend) reported `surface: bg` on a web chat. Verified empirically before and after — event went from `bg` to `ui`. Fixed in `525218dfd` with two regression tests.

apps/api/app/services/integrations/integration_inference_service.py
```diff
@@ -74,7 +75,10 @@ async def infer_integration_category(
     try:
         async with asyncio.timeout(_CATEGORY_INFERENCE_TIMEOUT_SECONDS):
             response = await ainvoke_llm(
-                get_helper_llm(), [HumanMessage(content=prompt)], label="integration_category"
+                get_helper_llm(),
+                [HumanMessage(content=prompt)],
+                label="integration_category",
+                feature=AIFeature.INTEGRATION_INFERENCE,
```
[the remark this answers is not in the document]
Vinitthakkar45: Valid — fixed in `0f12977e1`.

`infer_integration_category` now takes `user_id` keyword-only (matching its sibling `infer_integration_content`, which already did) and passes `config=metered_config(user_id)`; `publish_service` forwards the publisher id it already had.

Accountability: this is the same root cause as the `first_message_service` finding — I added the consumer that needs `user_id` and then only checked the call sites that had no user in scope, not the ones that had one and were not forwarding it. `publish_custom_integration` had the id three lines above the call.

Mutation-checked after the fix, and the first pass was not clean: `metered_config(user_id)` → `metered_config(None)` survived, because the test asserted the `label` kwarg but never the `config`. `test_returns_lowercased_valid_category` now pins `config["configurable"]["user_id"]`, and the module is clean.

apps/api/app/utils/research_utils.py
```diff
@@ -54,7 +55,10 @@ async def decompose_research_queries(
 
     try:
         response = await ainvoke_llm(
-            get_helper_llm(), [HumanMessage(content=prompt)], label="research_queries"
+            get_helper_llm(),
+            [HumanMessage(content=prompt)],
+            label="research_queries",
+            feature=AIFeature.RESEARCH,
```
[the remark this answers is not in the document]
Vinitthakkar45: Valid, and the cache warning was the useful half of this — fixed in `0f12977e1`.

`decompose_research_queries` now takes `user_id` keyword-only and passes `config=metered_config(user_id)`. The naive version of this change is a real regression: the function is `@Cacheable(smart_hash=True)`, and `create_cache_key_hash` folds in *every* argument, so simply adding the parameter would have turned one shared six-hour entry into one per user — every user paying again for an identical decomposition. It now uses an explicit `key_generator` that hashes the question and drops `user_id`:

```python
def _decompose_cache_key(func_name: str, *args: object, **kwargs: object) -> str:
    shared = {name: value for name, value in kwargs.items() if name != "user_id"}
    return f"research:{create_cache_key_hash(func_name, *args, **shared)}"
```

Three tests pin it — two users sharing one entry, a different question getting its own, and a non-`user_id` keyword still reaching the hash — and I verified the first goes red by leaking `user_id` back into the filter before trusting it.

Accountability: same root cause as the other two attribution findings (added the consumer, audited only the call sites with no user in scope). The cache-key hazard I would not have caught from the diff alone; it only shows up if you go read what `smart_hash` actually hashes. Worth noting for anyone adding metering to a cached helper later.

Mutation gate is clean on the module after adding assertions for the `config` and for `func_name`/kwargs reaching the hash — the `user_id` filter had made the original key test vacuous.

apps/api/app/services/onboarding/social_profile_service.py
```diff
@@ -283,7 +284,10 @@ async def extract_social_profiles_from_emails(
             candidates=candidates_text,
         )
         result: SocialProfileFilterOutput = await ainvoke_llm(
-            structured_llm, [HumanMessage(content=prompt)], label="onboarding_social_profile"
+            structured_llm,
+            [HumanMessage(content=prompt)],
+            label="onboarding_social_profile",
+            feature=AIFeature.ONBOARDING,
```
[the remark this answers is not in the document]
Vinitthakkar45: Valid finding, and **not fixed in this PR** — deliberately, and leaving this thread open so it is a visible decision rather than a silent gap.

The gap is real: `extract_social_profiles_from_emails` calls `ainvoke_llm` with no config, so `onboarding_social_profile` spend lands on nobody. Same root cause as the `first_message_service` and `integration_inference` findings, both of which are fixed here.

Why this one is not a one-liner. The function that needs the change *is* the complexity debt:

```
PLR0912 Too many branches (25 > 10)   social_profile_service.py:149
PLR0915 Too many statements (80 > 50) social_profile_service.py:149
```

Both are grandfathered in `tools/lints/plr_complexity_baseline.txt`, and that ratchet is touch-to-fix by design: the moment a PR modifies a listed file, its recorded violations must be fixed in the same PR. So adding `config=metered_config(user_id)` to line 285 makes the quality gate require splitting a 167-line function in the onboarding pipeline — an unrelated refactor with its own test surface, inside a PR whose scope is analytics attribution. Suppressing the rule instead is not an option we take.

Two smaller notes on the finding itself: `user_id` is not actually in scope in this function (it takes `user_name` and `user_email`), so it needs threading from `_run_social_profiles_background` as well; and the call is `ainvoke_llm` on a `with_structured_output` model, so it routes through the same auxiliary metering seam as the rest.

Until it is fixed, the failure is at least loud rather than invisible: `capture_auxiliary_llm_call` emits `log.warning("llm_call_unattributed", label="onboarding_social_profile", model=...)` on every one of these calls, so the missing attribution is greppable in the logs by label. Flagged to the PR author for the call on whether the split lands here or as its own change.

apps/api/app/services/llm_usage_analytics.py
```diff
@@ -0,0 +1,155 @@
...
+    and ``agent_name`` still carries the second half.
+    """
+    if workflow_id:
+        return AIFeature.WORKFLOW
+    if agent_name in TIER_AGENT_NAMES:
+        return AIFeature.CHAT
+    return AIFeature.INTEGRATION
+
+
+def feature_for_label(label: str) -> AIFeature:
```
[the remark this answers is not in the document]
Vinitthakkar45: Pushing back on this one — I think the guideline is being read more broadly than it is written.

`apps/api/CLAUDE.md` → "Service Layer" says: *"Services are async module-level functions, not classes. No service classes with `__init__`, instance methods, or injected dependencies."* The rule is about **hidden singletons and dependency injection**, not about the `async` keyword. Its own lint agrees: `tools/lints/no_service_classes.py` scopes itself to `*Service`-named classes carrying instance state, and states so explicitly — *"Scope is deliberately the `*Service` naming convention, not every class in `app/services`."* This module has no classes at all, so the lint has nothing to say about it. `tools/lints/run.py` is green: `6 rules passed on 919 files`.

Checked against the codebase rather than argued from the text alone:

- `app/services/analytics_service.py` — the module this one exists to extend, and the source of `capture_event` and `AnalyticsEvents` — is **nine sync module-level functions and zero async ones** (`identify_user`, `capture_event`, `capture_context_event`, `track_signup`, `track_login`, `track_logout`, `track_subscription_event`, `track_payment_event`, `_get_posthog_client`). PostHog capture is fire-and-forget into a batching client; there is nothing to await.
- `feature_for_label` is a pure dict lookup, and `capture_auxiliary_llm_call` calls a sync `capture_event`. Marking either `async` would add an await point with no I/O behind it — worse for the callers, which are already inside a metering hot path.

On the split itself: putting the taxonomy in a separate module would mean `LABEL_FEATURES` living apart from the one function that consumes it and the one event that emits it, in a 155-line module whose entire purpose is that mapping. The drift risk is exactly what keeping them in one file prevents — `tests/unit/services/test_llm_usage_analytics.py` has an AST test that walks `app/` for every `label=` passed to a metering helper and fails if the table has no entry for it, which is only coherent while the table and the resolver are one unit.

Happy to move it if the maintainers read the guideline as covering sync helpers generally, but that would also make `analytics_service.py` non-compliant, so it seems like the wrong reading. Leaving as-is.

apps/api/app/services/onboarding/first_message_service.py
```diff
@@ -36,21 +37,41 @@ def default_first_message(name: str) -> str:
...
+
+@dataclass(frozen=True)
+class FirstMessageOutcome:
+    """What onboarding actually produced, which is what the greeting reports."""
+
+    triage: InboxTriage | None
+    created_todos: list[OnboardingTodoSummary]
+    created_workflows: list[OnboardingWorkflowSummary]
+    executed_todos: list[OnboardingTodoSummary] | None = None
+    clarify_answers: list[ClarifyAnswerRecord] | None = None
```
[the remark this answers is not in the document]
Vinitthakkar45: Pushing back here too — same reading of the guideline, and the codebase is fairly clear on it.

`apps/api/CLAUDE.md` → "Service Layer" bans **service classes**: `__init__`, instance methods, injected dependencies — the hidden-singleton pattern, with `class TodoService: def __init__(self, db)` as its worked counter-example. `FirstMessageRecipient` and `FirstMessageOutcome` are frozen `@dataclass` argument bundles with no methods and no behaviour; they are not a service class in the sense the rule is about, and `tools/lints/no_service_classes.py` (which scopes to `*Service`-named stateful classes, and explicitly says so) passes.

Measured rather than asserted: **29 files under `app/services/` already define `@dataclass`**, including this feature's own direct neighbours — `app/services/llm_metering.py`, `app/services/hil/gate.py`, `app/services/chat/stream.py`, `app/services/workflow/playbook/runner.py`, `app/services/mcp/mcp_client_pool.py`. A local argument-bundle beside its only consumer is the established convention here, not a deviation from it.

On why they exist at all: `generate_first_message` had reached 11 parameters and tripped `PLR0913` once the complexity ratchet started applying to this file. The two dataclasses group the arguments along the seam they already had (who the message is for, what the pipeline found), which is the same shape `client.py` uses for `LLMInvokeOptions` and `agent_helpers.py` uses for `AgentTurn` / `AgentIdentity`. Moving them to `app/models/` would put a private, single-consumer argument shape in the shared model namespace and invite reuse of something that is deliberately local — the opposite trade to the one the guideline is protecting.

Leaving as-is. If the convention should be "no dataclasses under `app/services/`" then that is a 29-file change and a `CLAUDE.md` amendment, not something to settle inside this PR.

apps/api/app/utils/research_utils.py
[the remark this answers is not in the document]
Vinitthakkar45: Reopening this and **reverting the fix** — I got the scope call wrong the first time, so here is the full trace.

The finding is valid and my fix worked. What I underestimated was what threading `user_id` through `deep_research` costs. `research_tool.py` carries a grandfathered `PLR0915` (72 statements > 50), and the complexity ratchet is touch-to-fix, so editing one line to forward `user_id` made that violation due. Satisfying it meant splitting the function — which I did, cleanly, into `_search_sub_queries` and `_fetch_ranked_sources`. But moving ~170 lines of fetch pipeline puts all of it on the PR's changed lines, and the mutation gate then correctly demanded roughly **70 new assertions** on crawl4ai timeouts, tier error prefixes (`"crawl4ai: "` / `"httpx: "`), semaphore counts and progress strings. None of that is analytics work, and much of it would be brittle string assertions.

So this PR now touches neither file, and the two attribution fixes that survive are the high-volume ones: onboarding first-message generation and integration category inference. `research_queries` fires once per `deep_research` invocation on a helper model — the smallest of the three gaps by a wide margin.

This is the identical cascade as the `social_profile_service` thread, and I am declining it for the identical reason rather than fixing one and hand-waving the other. Both threads stay open as visible decisions.

Worth keeping from the reverted work, for whoever picks this up: the naive fix is a **latent regression**, not just a missing config. `decompose_research_queries` is `@Cacheable(smart_hash=True)`, and `create_cache_key_hash` folds in every argument — so simply adding the parameter turns one shared six-hour entry into one per user, and every user re-pays for an identical decomposition. It needs an explicit `key_generator` that drops `user_id` from the hash. Thanks for flagging that half in the original comment; it was the part I would not have caught from the diff alone.

Not silent in the meantime: `capture_auxiliary_llm_call` logs `llm_call_unattributed` with `label="research_queries"` on every such call, so the gap is greppable rather than invisible.

apps/api/app/services/onboarding/first_message_service.py
```diff
@@ -36,21 +37,41 @@ def default_first_message(name: str) -> str:
...
-    writing_style: WritingStyleProfile | None,
-    has_gmail: bool,
-    focus: str = "",
-    executed_todos: list[OnboardingTodoSummary] | None = None,
-    clarify_answers: list[ClarifyAnswerRecord] | None = None,
+    recipient: FirstMessageRecipient,
+    outcome: FirstMessageOutcome,
 ) -> str:
     """Generate GAIA's first message to a new user."""
+    user_id, name = recipient.user_id, recipient.name
```
aryanranderiya, line 68: why is this needed? seems unnecssary

apps/api/app/services/integrations/integration_inference_service.py
```diff
@@ -58,6 +58,8 @@ async def infer_integration_category(
     description: str,
     tools: list[dict[str, Any]],
     server_url: str,
+    *,
```
aryanranderiya, line 61: why are we using *

apps/api/app/services/onboarding/first_message_service.py
```diff
@@ -36,21 +37,41 @@ def default_first_message(name: str) -> str:
     )
 
 
+@dataclass(frozen=True)
+class FirstMessageRecipient:
```
aryanranderiya, line 41: wrong place to define a type

apps/api/app/services/analytics_service.py
```diff
@@ -146,6 +146,39 @@ class AnalyticsEvents(StrEnum):
 
     USAGE_QUERIED = "usage:queried"
 
+    # Background LLM spend only. Agent-graph calls are already covered by
+    # PostHog's own $ai_generation, so this never duplicates them.
+    AI_LLM_CALL_COMPLETED = "ai:llm_call_completed"
+
+
+class AIFeature(StrEnum):
```
aryanranderiya, line 154: why is this needed? surely we have something like this

apps/api/app/services/llm_usage_analytics.py
```diff
@@ -0,0 +1,155 @@
...
+TIER_AGENT_NAMES = frozenset({"comms_agent", "executor_agent"})
+
+#: Which capability each auxiliary ``label`` belongs to.
+#:
+#: Keyed on the label every one-shot already passes for its log line, rather
+#: than a second argument at each call site: one table is the whole taxonomy,
+#: readable at a glance, and adding a helper never means editing a service file
+#: for an analytics reason. An unmapped label is not silent — it books to
+#: ``UNATTRIBUTED`` and logs an error naming itself.
+LABEL_FEATURES: dict[str, AIFeature] = {
```
aryanranderiya, line 35: why can't this be a label attribute on the aifeature.xyz instead?

aryanranderiya (changes requested):

apps/api/app/services/onboarding/first_message_service.py
Vinitthakkar45: Done — both moved to `app/models/onboarding_models.py`, next to the types they are built from.

Left them as dataclasses rather than Pydantic models: a few files in that folder already do this, and these never go over the network so there is nothing to validate.

apps/api/app/services/onboarding/first_message_service.py
Vinitthakkar45: The repo has a check on how many arguments a function can take. This file already failed it before I touched it, and was sitting on a "known, fix later" list. That exemption ends the moment anyone edits the file — and my one line here (charging this call to a user) counts as editing it.

Grouping the arguments is how I got it passing.

Which do you want?
1. Leave it as is.
2. I undo the grouping and put the file back on the ignore list.

I would go with 1, because 2 means that list never shrinks. Reply with a number and I will do it.

apps/api/app/services/integrations/integration_inference_service.py
Vinitthakkar45: (Answered below — removing it.)

apps/api/app/services/analytics_service.py
Vinitthakkar45: Yes, `rate_limits.py` has a feature list and about a third of the names look the same.

They answer different questions though. That one is "how much of this is the user allowed to do" — which is why it holds things like download and weather checks. Mine is "which part of the product spent money on an AI call", including background work the user never asked for: deciding whether to ask permission, naming a chat, pulling text out of a file. None of those have limits, and putting them in the list that gates what people can do would be wrong.

Which do you want?
1. Keep the two lists separate (what the PR does now).
2. I extend the rate-limit list to cover the internal calls too, and delete mine.

I would go with 1. Reply with a number and I will do it in this PR.

apps/api/app/services/llm_usage_analytics.py
Vinitthakkar45: Done. The feature list and the table that maps each call to a feature now sit together in one file, so you cannot change one and miss the other.

I kept the table as a dictionary instead of putting the names on the enum. The lookup runs call name -> feature, which a dictionary answers directly; your way stores it backwards and needs flipping at startup.

Being next to each other does not enforce anything on its own, so there is also a test that fails if a feature has no way of ever being produced. It found one called IMAGE that nothing could create. Deleted.

Nothing needed from you here unless you still want it on the enum — say so and I will switch it.

apps/api/app/services/integrations/integration_inference_service.py
Vinitthakkar45: Removed.

Also took it off the other function in this file so the two match, and off one function in `llm_usage_analytics.py`. Nothing else had to change, because every caller was already passing the argument by name anyway.

apps/api/app/services/llm_usage_analytics.py
aryanranderiya: let's consolidate this, separate lists will keep drifting apart and 10 months down the line nobody will remember which list is for what

aryanranderiya (changes requested):

apps/api/app/services/llm_usage_analytics.py
Vinitthakkar45: Consolidated in `d1239101f` — the dictionary is gone.

Each member now carries its own labels:

```python
class AIFeature(StrEnum):
    CHAT = "chat"                                    # graph spend, no labels
    HIL = "hil", ("hil_intent_judge", "hil_tool_classification", ...)
    ONBOARDING = "onboarding", ("onboarding_clarify", ...)
```

`AIFeature.for_label()` answers the lookup from an index built off the members, so there is no second list left to drift.

One thing I checked rather than assumed: a custom `__new__` on a `StrEnum` can quietly change how members render. `.value` is still the plain string, f-strings still print `onboarding` and not `AIFeature.ONBOARDING`, and members still work as plain-string dict keys — so every event property and log line keeps its shape.

Two tests hold the line, both watched failing first: one fails if a member has no way to be produced, the other if a member claims a label no call site passes.

PR #1196, opened 2026-09-04
feat(agents): integration-activation executor mode behind a flag

Dhruv-Maradiya:
## Summary

Adds an **integration-activation** mode for the executor agent behind a new `ENABLE_INTEGRATION_ACTIVATION` flag (default **off**). Instead of handing work off to a per-integration subagent graph, the executor calls `activate_integration(integration_id)` to pull that integration into its own context — its tools registered and the most-used ones bound in the same turn, its operating prompt, which account the user is on, their standing instructions, and its skills — then does the work itself. Delegation, when needed, goes through the generic `spawn_subagent`, which inherits the bound tools.

When the flag is off, the existing `handoff` / per-integration-subagent path is unchanged.

## What's in here

- **`activate_integration` tool** (`integration_activation.py`) + `register_integration_tools` extracted from `create_subagent` as the single place an integration's tools become bindable (shared by provider subagents and activation).
- **Activation-variant executor prompt** (`executor_activation_prompt.py`) — rewrites the handoff-specific passages of `EXECUTOR_AGENT_PROMPT`, anchored to that text so the two variants raise rather than drift silently.
- **Shared provider-metadata block** between the subagent context section and activation, so an executor acting on an integration directly knows which account it's on.
- **Lean subagent spawn** — a spawned subagent no longer inherits the parent's bound tool set; it binds its own minimal operational set and retrieves the rest on demand, avoiding carrying ~14k of activated-integration schemas.
- **Compaction unwraps `Command`-wrapped tool outputs** so large payloads from tool-binding tools (`spawn_subagent`, `activate_integration`) still get shed, with the `Command`'s `goto`/`resume` preserved.
- **Dev custom LLM lane** runs `ChatOpenAI` (the OpenRouter SDK rejects OpenAI-compatible envelopes such as OpenCode Zen's) with a browser UA for Cloudflare-fronted lanes; `CUSTOM` dropped from sticky routing.
- **Dev harness** honors a per-run `model` key (`DEV_MODEL_OPTIONS`); adds `agent_bench.py` / `doctrine_bench.py` benchmarking scripts.

## Tests

New unit tests for activation, the activation prompt builder, graph-builder wiring, provider-subagent tool registration, and compaction Command-unwrapping; existing subagent/context tests updated.

## Verification

- [ ] `nx lint api` / `nx type-check api`
- [ ] `cd apps/api && uv run pytest` for the touched suites

> Base branch is **`master-dhruv`**, not `master`.


Dhruv-Maradiya:
@coderabbitai review

PR #1204, opened 2026-09-07
fix(onboarding): surface personalization save failures instead of logging outcome=ok

aryanranderiya:
## What

`_run_holo_card` reported `outcome=ok` for a personalization save that never happened.

The call went through `post_onboarding_service.save_personalization_data`, a pass-through wrapper around `user_repository.save_personalization` that caught its own exception, logged it, and returned `None`. `_run_holo_card` wrapped the same call in its own `try/except`, but nothing ever reached it — so a failed Mongo write produced:

```
[ONBOARDING] Error saving personalization data   <- from the wrapper
[ONBOARDING] holo_card done, outcome=ok          <- from the caller, 1ms later
```

The `outcome` field is the one an operator filters on. It said the card was saved.

Fixed at the root rather than at the symptom: the wrapper is deleted and `_run_holo_card` calls `user_repository.save_personalization` directly, so a raising write lands in the node's own `except` and produces `outcome=failed` with the error type. `HOLO_READY` is still emitted either way — the frontend waits on it, and skipping it leaves the card spinning.

`post_onboarding_service.py` now holds only `seed_initial_user_data`; its module docstring was updated to match.

## Test debt paid in the same PR

Touching `apps/api/tests/e2e/test_onboarding.py` un-grandfathers its two PLR entries, so both are fixed and their baseline lines deleted:

- **PLR0915** — the 72-statement `world` fixture is split into five composed fixtures (`persistence_patches`, `composio_patches`, `gmail_patches`, `llm_patches`, `downstream_patches`), assembled by `all_patches` and entered by a four-line `world`. Same patches, same order, same assertions.
- **PLR0913** — the 8-parameter `_fetch_emails` stub had drifted from the real `fetch_emails_for_onboarding`, which folded `fmt`/`include_sent` into `OnboardingFetchOptions`. The stub now mirrors the real signature (6 params), so it also stops being a stale fixture that could pass while the real call site broke.

## Verified

Red first — with the product code still unfixed, the new test failed on exactly the misleading field:

```
$ uv run --directory apps/api pytest -p no:xdist -o addopts="" \
    tests/unit/services/onboarding/test_intelligence_service_creation.py -q -k save_failure
>       assert [c.kwargs.get("outcome") for c in log.info.call_args_list] == []
E       AssertionError: assert ['ok'] == []
1 failed, 60 deselected in 2.90s
```

After the fix:

```
$ uv run --directory apps/api pytest -p no:xdist -o addopts="" \
    tests/unit/services/onboarding tests/unit/services/test_onboarding_service.py -q
305 passed in 4.88s

$ uv run --directory apps/api pytest -p no:xdist -o addopts="" tests/e2e/test_onboarding.py -q
83 passed in 3.12s      # 83 before the fixture split as well

$ uv run --directory apps/api --group dev ruff check .
All checks passed!

$ uv run --directory apps/api --group backend --group dev mypy app --ignore-missing-imports
Success: no issues found in 917 source files

$ python3 tools/lints/check_plr_complexity.py
plr-complexity-ratchet: 148 current violation(s), all grandfathered and untouched
```

The e2e file runs on fakeredis with every external patched, so no Docker services were needed.

## Not verified

- The onboarding pipeline was **not** driven against a live stack — no real Mongo write was made to fail and no real `outcome=failed` log line was observed. The proof is the unit test at the `user_repository.save_personalization` seam.
- Only the two onboarding test directories above were run, not the full API suite; CI covers the rest.

## Adjacent debt, surfaced not fixed

`seed_initial_user_data` in the same file still catches its own exception and returns normally. Its caller does not branch on the result, so there is no misleading `outcome` to correct today — but it is the same shape as the bug above. Flagging rather than folding it in, since fixing it means deciding what the caller should do on a failed seed.

### DIFF
--- a/apps/api/app/services/onboarding/intelligence_service.py
+++ b/apps/api/app/services/onboarding/intelligence_service.py
@@ -90,9 +90,6 @@ from app.services.onboarding.first_message_service import (
     generate_first_message,
 )
 from app.services.onboarding.inbox_triage_service import triage_inbox
-from app.services.onboarding.post_onboarding_service import (
-    save_personalization_data,
-)
 from app.services.onboarding.social_profile_service import (
     dedup_profiles_by_platform,
     extract_social_profiles_from_emails,
@@ -1078,17 +1075,17 @@ async def _run_holo_card(
         )
         phrase_bio_duration_s = round(time.monotonic() - t_phrase_bio, 2)
         t_save = time.monotonic()
-        await save_personalization_data(
+        await user_repository.save_personalization(
             ctx.user_id,
-            card_design.house,
-            phrase,
-            user_bio,
-            bio_status,
-            [],
-            metadata.account_number,
-            metadata.member_since,
-            card_design.overlay_color,
-            card_design.overlay_opacity,
+            house=card_design.house,
+            personality_phrase=phrase,
+            user_bio=user_bio,
+            bio_status=bio_status,
+            account_number=metadata.account_number,
+            member_since=metadata.member_since,
+            overlay_color=card_design.overlay_color,
+            overlay_opacity=card_design.overlay_opacity,
+            workflow_ids=[],
         )
         log.info(
             f"{LogTag.ONBOARDING} holo_card done",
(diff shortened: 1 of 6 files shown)


aryanranderiya:
Folded into #1203 (merged as a03ad21d93).

PR #1217, opened 2026-09-12
ci: probe run for #1213 head (do not merge)

aryanranderiya:
Throwaway probe: GitHub stopped computing mergeability for #1213 after its first push, so its pull_request workflows never fire. This PR exists only to run CI on the same head commit (d838aadd3). Will be closed.


aryanranderiya:
Probe done: it proved the branch computes and runs CI on a fresh PR; #1218 is the real one.

PR #1218, opened 2026-09-12
feat(api): one error envelope, response models everywhere, generated TS types with a drift gate

aryanranderiya:
## What

The API's types become the contract, and the frontends consume it instead of re-typing it.

- **One error envelope.** `ErrorEnvelope` (`apps/api/app/schemas/errors.py`) is the body of every non-2xx response — `AppError`, `HTTPException`, validation, crashes, and the four middlewares that used to answer with their own `{detail}`. Declared on every router and on every route-level `responses=` for 4xx/5xx/422; `tests/meta/test_route_contract.py` asserts every documented error of every route is the envelope. Web/shared/bots/mobile/CLI read one shape.
- **Every route has a `response_model`**, no bare `Any` bodies, operation ids are `<tag>_<name>` (`app/core/openapi.py`); the route-contract ratchet keeps it so. Every model has a unique class name — `scripts/export_openapi.py` refuses a collision instead of letting FastAPI mangle it into `app__models__…`.
- **Generated types, exported by name.** `apps/api/scripts/export_openapi.py` → `apps/api/openapi.json` → `openapi-typescript --root-types` → `libs/shared/ts/src/api/generated/schema.d.ts`. `import type { TodoResponse } from "@gaia/shared/api/generated"` — the Pydantic class name, navigable and greppable; a dual-use model comes as `NameInput`/`NameOutput`. `mise api:types` runs it; output is byte-deterministic.
- **Gates.** prek hook regenerates on commit; CI lane `api-schema` fails on drift; `api-schema-types` fails any hand-written twin of a schema type (the only allowed forms are a re-export or an alias of a generated name) and any raw `apiService` call in web feature code.
- **Migration complete.** Every hand-written API type across web, shared, mobile, bots and the CLI is gone; web calls go through the path-typed client `apps/web/src/lib/api/typed.ts`. Rules in `.claude/rules/general.md`, `apps/web/CLAUDE.md`, `apps/api/CLAUDE.md`.

## Bugs the types found (and fixed here)

- api: `GmailMessageSummary` rejected the nulls Composio documents (`threadId`, `subject`, headers) and a bodiless Gmail-API message — a 500 on the mail routes and a failed writing-style pass in onboarding; the transform now emits the declared shape. `PollPairingResponse.status` was a bare `str` the CLI switched on with a state the API never emits. A marketplace row without a slug (explore-only, or a backfill that ran out of retries) 500ed the whole list; `is_explore` was queried and seeded but never declared on the model. `ErrorEnvelope` could 500 on a dict `detail` without `message` or a `meta` carrying `message`/non-str `code`. `page`/`search` on `GET /triggers/options` were scraped off the raw query string, so the typed client could not send them and the GitHub repo picker silently lost paging — they are declared now.
- web: four client modules had an import hoisted above `"use client"`, which broke `next build` (type-check cannot see it).
- mobile: image `improvedPrompt` vs `improved_prompt`; dashboard `total_count` never sent (and `?activated=true` ignored by the route); a workflow polling hook waited on progress fields the server never writes and had no caller; discover previews rendered as installed skills; skill `category` never sent; nonexistent `scheduled` trigger type; a hand-written push-token pair.
- cli: still read `{detail}` on errors; hand-written twins of the three device-pairing responses renamed to dodge the gate.
- api: `WorkflowResponse` list endpoint returned the wrong view; `UserSubscriptionStatus` carried untyped dicts; the OAuth callbacks' business logic lived in the route files.

## Verified

`checks.mjs api-schema` ✅ · `api-schema-types` ✅ · route-contract ratchet 6/6 ✅ · mypy 946 files ✅ · ruff ✅ · hermetic API suite 18,971 passed (the two `test_webhook_composio_pipeline` cases need the real broker and pass on CI) · tsc: all 12 workspaces 0 errors · web vitest 70 files / shared 32 / harness tooling 477 ✅ · `next build` ✅ · mutation gate: 474 survivors killed, 0 gap lines (verified per module with `scripts/ci/mutation.sh local`) · adversarial contract audit (six findings, all fixed above).

## Stack

Stacked on #1175 (`master` ← #1079 ← #1161 ← #1202 ← #1175 ← this). Supersedes #1213: GitHub stopped computing that PR's mergeability after its first push, so its `pull_request` workflows never fired again; this PR is the same branch.

https://claude.ai/code/session_01LVTV88ezWxscioR7gvaV7j


aryanranderiya:
@greptileai re-review

aryanranderiya:
@greptileai re-review

aryanranderiya:
@greptileai review

aryanranderiya:
@coderabbitai review

aryanranderiya:
@greptileai review

